PR #3, opened 2020-01-09
Add `#![no_std]` to the reference impl

jonas-schievink:
This ensures that no accidental uses of libstd can sneak in.

I've verified that this now compiles for ARMv6-M targets via `cargo build --target thumbv6m-none-eabi`.

(this is great work btw!)

### DIFF
--- a/reference_impl/reference_impl.rs
+++ b/reference_impl/reference_impl.rs
@@ -1,3 +1,5 @@
+#![no_std]
+
 use core::cmp::min;
 use core::convert::TryInto;
 


oconnor663:
I've added the `#![no_std]` directive temporarily to make sure it works, but I decided against checking it in. The reference impl isn't published on crates.io, so there won't be any real `no_std` callers depending on it. (The production `blake3` crate does support `no_std` if you use `--no-default-features`.) And it's not expected to receive many changes going forward.

The main purpose of the reference impl (besides testing) is to help folks understand the algorithm, together with the spec. I expect that a lot of the people reading it won't be experienced Rust coders, so I've tried to minimize the use of advanced Rust features. I think the benefit of including `#![no_std]` might be smaller than the cost of making people wonder what it does.

Issue #8, opened 2020-01-10
Callable from C (no_mangle pub extern as in c_neon)


rurban:
Would it be possible to export the fast mt versions from Rust to C/C++, to compare against the single-threaded C version? I'm pretty sure many C/C++ projects would want to use the faster rust versions.

I'd also want to compare the C versions against the rust versions in smhasher.

oconnor663:
I'm leaning towards experimenting with this as a separate crate at first. If we find it's getting some users, and merging the two repos would simplify maintenance, maybe we could do that in the future?

My instinct is that most C users are going to strongly prefer the pure C version, to avoid taking a dependency on `rustc`. We also plan to spruce that version up in the coming weeks, and the performance gap between the two might close at some point.

PR #9, opened 2020-01-10
Add staticlib, callable from C

rurban:
See GH #8
(Untested but looks fine, maybe a C header file `blake3_rs.h` can be added also)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,10 @@ c_avx512 = []
 c_neon = []
 std = []
 
+[lib]
+name = "blake3"
+crate-type = ["staticlib"]
+
 [dependencies]
 arrayref = "0.3.5"
 arrayvec = { version = "0.5.1", default-features = false, features = ["array-sizes-33-128"] }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -673,6 +673,12 @@ pub fn keyed_hash(key: &[u8; KEY_LEN], input: &[u8]) -> Hash {
     hash_all_at_once(input, &key_words, KEYED_HASH).root_hash()
 }
 
+#[no_mangle]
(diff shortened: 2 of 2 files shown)


elichai:
I'd replace the inputs with raw pointers (C) and at least asserts for nullness, or even return int like this: https://github.com/BLAKE2/BLAKE2/blob/ca4c89314abff54e3806b44e4a08164f8204f09a/sse/blake2sp.c#L78:L84

oconnor663:
C bindings should probably expose the incremental API. But this is tricky, because the `Hasher` type is not currently `repr(C)`, and we don't want to force C to dynamically allocate it behind an opaque pointer. Maybe we could make it `repr(C)`? Or maybe it would make more sense to expose some more of the internal guts in a semi-stable way, by expanding https://github.com/BLAKE3-team/BLAKE3/blob/master/src/guts.rs.

oconnor663:
I think the better approach here will be to just improve the C implementation to match the Rust one.

cemeyer:
I like the idea of leaving the multithreaded, crypto-sensitive code in memory-safe rust and just linking the library as a build artifact from C code, but I can certainly see the appeal of a pure-C version as well. 
 So long as it has both feature parity and, uh, no bugs.

oconnor663:
Note that there's a fair bit of unsafe code in the Rust version, mainly because using SIMD is inherently unsafe. It's possible that this could be reduced with some more abstractions, but at the same time there's a lot of value in being able to see exactly what intrinsics are used where, and I'm not eager to make the implementations more abstract.

cemeyer:
Sure, that's reasonable.

oconnor663:
The C implementation is now on par with the Rust implementation in terms of single-threaded performance. I'm going to close this for now.

PR #10, opened 2020-01-10
Fix misspelled words_from_little_endian_bytes

leobuskin:


oconnor663:
Oh dear, thank you for catching this >.<

PR #12, opened 2020-01-10
Update license field in Cargo.toml

bbqsrc:
Fixes the issue of crates.io not recognising the license, and using the `OR` terminology to describe the dual licensing option.


oconnor663:
Thanks you, I'll land this and then make the same change in `b3sum/Cargo.toml`.

PR #14, opened 2020-01-11
Inline trivial functions

cesarb:
For the Read and Write traits, this also allows the compiler to see that
the return value is always Ok, allowing it to remove the Err case from
the caller as dead code.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -113,10 +113,12 @@ const KEYED_HASH: u8 = 1 << 4;
 const DERIVE_KEY_CONTEXT: u8 = 1 << 5;
 const DERIVE_KEY_MATERIAL: u8 = 1 << 6;
 
+#[inline]
 fn counter_low(counter: u64) -> u32 {
     counter as u32
 }
 
+#[inline]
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -409,6 +417,7 @@ fn left_len(content_len: usize) -> usize {
 // right side off to another thread, if idle threads are available. If the
 // "rayon" feature is disabled, just make ordinary function calls for the left
 // and the right.
+#[inline]
```
oconnor663: Out of curiosity, are templated functions always effectively inline?

cesarb: > Out of curiosity, are templated functions always effectively inline?

No, they're only monomorphized. However, both gcc and llvm love to inline functions which are only called in a single place, and this function is only called in a single place. Even if later it's called in more than one place, it might still be inlined if LLVM decides it's too small. The only place where `#[inline]` is really mandatory is for cross-crate inlining (at least for now, perhaps in the future rustc might detect small functions at the MIR level and automatically make them inline).

PR #17, opened 2020-01-11
Use fixed-size constant_time_eq

cesarb:
The generic constant_time_eq has several branches on the slice length,
which are not necessary when the slice length is known. However, the
optimizer is not allowed to look into the core of constant_time_eq, so
these branches cannot be elided.

Use instead a fixed-size variant of constant_time_eq, which has no
branches since the length is known.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ std = []
 [dependencies]
 arrayref = "0.3.5"
 arrayvec = { version = "0.5.1", default-features = false, features = ["array-sizes-33-128"] }
-constant_time_eq = "0.1.4"
+constant_time_eq = "0.1.5"
 # A performance note for the "rayon" feature: Multi-threading can have
 # significant overhead for small inputs, particularly on x86 where individual
 # cores are very fast. On the other hand, on slower platforms like ARM,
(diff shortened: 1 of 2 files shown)


oconnor663:
Very snazzy! What prompted you to add these?

cesarb:
> Very snazzy! What prompted you to add these?

Looking at the generated assembly in my new 64-bit phone (compiled within Termux). The variable-size `constant_time_eq` does a comparison between both sizes (for the assert), a comparison of the size with zero (to return immediately), a comparison of the size with 31 (to skip the vectorized code), an AND with a mask, the vectorized XOR/OR code, and a conventional XOR/OR loop for the tail. There's a total of 7 labels in that function (though all of the branches should be easy for the branch predictor when always called with the same size). The constant-size `constant_time_eq`, on the other hand, has no branches, and should have only the vectorized XOR/OR part.

Unfortunately, the code generated by LLVM (at least on the version Termux has) for the fixed-size variants on this phone is atrocious; it fully unrolls both the 16-byte and the 32-byte variants, somehow managing to spill several registers to the stack. The code for the 64-bit variant, on the other hand, is a fine vectorized code, as is the code for all three variants on x86 (at least on the version Fedora 31 has). Since this is probably an optimization bug in LLVM for aarch64 (it should at least do it without spilling), it should get better with time.

Issue #18, opened 2020-01-12
Allow b3sum to output raw


tom-adsfund:
Please could you provide the option to output the hash as the raw bytes?

oconnor663:
This sounds like a good idea. Probably it would be an error if you supplied more than one argument in this case? Could you tell me more about your use case?

tom-adsfund:
The idea is that I want to create a large hash and save that to a file, and yet don't want to keep it in hex format. I don't really want to go into any extra details!

phayes:
I've added PR #22 to resolve this issue.  

Raw output is super useful.  I've found myself needing it for piping the raw digest over to another process like openssl for signing.

oconnor663:
Thank you! Commented on the PR.

Trendsetter74:
I wish I knew more about what's going on but I'm very simple person just
need to learn how to code

Kenneth Smith

PR #22, opened 2020-01-13
Adds support for raw output to b3sum

phayes:
This PR adds support for raw output (as opposed to hex encoded output). 

This is useful for piping the output of the hash function to another process (eg to openssl for signing). 

```
b3sum --raw README.md
����.2O����c��+�=[F�˱��STQ
```

Some details:
 1. Raw output only works on stdin or a single file. It will return an error if multiple files are passed. This is documented in `--help`
2. No file names are displayed with `--raw`
3. No terminating linebreak is printed. This is correct behavior for raw binary output.

This resolve issue #18 .

### DIFF
--- a/b3sum/src/main.rs
+++ b/b3sum/src/main.rs
@@ -191,28 +191,26 @@ fn main() -> Result<()> {
 
     if let Some(files) = args.values_of_os(FILE_ARG) {
         if raw_output && files.len() > 1 {
-            did_error = true;
-            eprintln!("b3sum: Only one filename can be provided when using --raw");
-        } else {
-            for filepath in files {
-                let filepath_str = filepath.to_string_lossy();
-                match hash_file(&base_hasher, filepath) {
-                    Ok(output) => {
-                        if raw_output {
-                            write_raw_output(output, len)?;
+            bail!("b3sum: Only one filename can be provided when using --raw");
+        }
+        for filepath in files {
+            let filepath_str = filepath.to_string_lossy();
+            match hash_file(&base_hasher, filepath) {
+                Ok(output) => {
+                    if raw_output {
+                        write_raw_output(output, len)?;
+                    } else {
+                        write_hex_output(output, len)?;
+                        if print_names {
+                            println!("  {}", filepath_str);
                         } else {
-                            write_hex_output(output, len)?;
-                            if print_names {
-                                println!("  {}", filepath_str);
-                            } else {
-                                println!();
-                            }
+                            println!();
                         }
                     }
-                    Err(e) => {
-                        did_error = true;
-                        eprintln!("b3sum: {}: {}", filepath_str, e);
-                    }
+                }
+                Err(e) => {
+                    did_error = true;
+                    eprintln!("b3sum: {}: {}", filepath_str, e);
                 }
             }
         }
--- a/b3sum/src/main.rs
+++ b/b3sum/src/main.rs
@@ -10,6 +10,7 @@ const LENGTH_ARG: &str = "length";
 const KEYED_ARG: &str = "keyed";
 const DERIVE_KEY_ARG: &str = "derive-key";
 const NO_NAMES_ARG: &str = "no-names";
+const RAW_ARG: &str = "raw";
 
 fn clap_parse_argv() -> clap::ArgMatches<'static> {
     App::new("b3sum")
@@ -41,6 +42,11 @@ fn clap_parse_argv() -> clap::ArgMatches<'static> {
             Arg::with_name(NO_NAMES_ARG)
                 .long(NO_NAMES_ARG)
                 .help("Omits filenames in the output"),
+        )
+        .arg(
+            Arg::with_name(RAW_ARG)
+                .long(RAW_ARG)
+                .help("Raw output without hexidecimal encoding. \nOnly one filename may be provided and no filenames will be displayed."),
         )
         .get_matches()
 }
@@ -105,7 +111,7 @@ fn maybe_hash_memmap(
     Ok(None)
 }
 
-fn write_output(mut output: blake3::OutputReader, mut len: u64) -> Result<()> {
+fn write_hex_output(mut output: blake3::OutputReader, mut len: u64) -> Result<()> {
     // Encoding multiples of the block size is most efficient.
     let mut block = [0; blake3::BLOCK_LEN];
     while len > 0 {
@@ -115,6 +121,19 @@ fn write_output(mut output: blake3::OutputReader, mut len: u64) -> Result<()> {
         print!("{}", &hex_str[..2 * take_bytes as usize]);
         len -= take_bytes;
     }
+    Ok(())
+}
+
+fn write_raw_output(mut output: blake3::OutputReader, mut len: u64) -> Result<()> {
+    let mut stdout = std::io::stdout();
+    let mut block = [0; blake3::BLOCK_LEN];
+    while len > 0 {
+        output.fill(&mut block);
+        let take_bytes = cmp::min(len, block.len() as u64);
+        stdout.write(&block[..take_bytes as usize])?;
+        len -= take_bytes;
+    }
+
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


phayes:
Oh, I also snuck another little change in here.  All errors are now printed to stderr instead of stdout.

b3sum/src/main.rs
```diff
@@ -118,6 +124,19 @@ fn write_output(mut output: blake3::OutputReader, mut len: u64) -> Result<()> {
     Ok(())
 }
 
+fn write_raw_output(mut output: blake3::OutputReader, mut len: u64) -> Result<()> {
+    let mut stdout = std::io::stdout();
```
oconnor663: We should probably call `lock()` here, since we're about to use this in a loop.

b3sum/src/main.rs
```diff
@@ -118,6 +124,19 @@ fn write_output(mut output: blake3::OutputReader, mut len: u64) -> Result<()> {
...
+    let mut block = [0; blake3::BLOCK_LEN];
+    while len > 0 {
+        output.fill(&mut block);
+        let take_bytes = cmp::min(len, block.len() as u64);
+        stdout.write(&block[..take_bytes as usize])?;
+        len -= take_bytes;
+    }
```
oconnor663: Rather than this loop, can we just use `std::io::copy`? Maybe together with `Read::take`? (A call to `lock()` above is probably still a good idea.)

b3sum/src/main.rs
```diff
@@ -166,31 +185,46 @@ fn main() -> Result<()> {
...
-        for filepath in files {
-            let filepath_str = filepath.to_string_lossy();
-            match hash_file(&base_hasher, filepath) {
-                Ok(output) => {
-                    write_output(output, len)?;
-                    if print_names {
-                        println!("  {}", filepath_str);
-                    } else {
-                        println!();
+        if raw_output && files.len() > 1 {
```
oconnor663: Rather than indenting the main loop by another level, let's `bail!` if this fails.

b3sum/src/main.rs
```diff
@@ -42,6 +43,11 @@ fn clap_parse_argv() -> clap::ArgMatches<'static> {
                 .long(NO_NAMES_ARG)
                 .help("Omits filenames in the output"),
         )
+        .arg(
+            Arg::with_name(RAW_ARG)
+                .long(RAW_ARG)
+                .help("Raw output without hexidecimal encoding. \nOnly one filename may be provided and no filenames will be displayed."),
```
oconnor663: Interesting, I didn't know Clap did the right thing with newlines in these strings. Very useful!

b3sum/tests/test.rs
```diff
@@ -93,3 +105,11 @@ fn test_length_without_value_is_an_error() {
         .run();
     assert!(result.is_err());
 }
+
+#[test]
+fn test_raw_with_multi_files_is_an_error() {
+    let result = cmd!(b3sum_exe(), "--raw", "file1", "file2")
```
oconnor663: I think this test could fail to test what it looks like, because the fact that `file1` and `file2` don't exist will lead to another type of error. Could you create a couple temporary files (with `NamedTempFile` as in other tests) to prevent this error from happening? You should be able to confirm that if you comment out the `len() > 1` check in main, that this test no longer gets and error and starts failing.

b3sum/tests/test.rs
```diff
@@ -11,7 +11,19 @@ pub fn b3sum_exe() -> PathBuf {
...
-    assert_eq!(&*expected, &*output);
+    assert_eq!(&*expected, output);
+}
+
+#[test]
+fn test_hash_one_raw() {
+    let expected = blake3::hash(b"foo").as_bytes().to_owned();
+    let mut stdout = Vec::new();
+    let mut output_reader = cmd!(b3sum_exe(), "--raw")
+        .stdin_bytes("foo")
```
oconnor663: You could also use `stdout_capture` here, if you wanted to save yourself the `read_to_end` call. Either way.

b3sum/src/main.rs
```diff
@@ -166,31 +185,46 @@ fn main() -> Result<()> {
...
+                            if print_names {
+                                println!("  {}", filepath_str);
+                            } else {
+                                println!();
+                            }
+                        }
+                    }
+                    Err(e) => {
+                        did_error = true;
+                        eprintln!("b3sum: {}: {}", filepath_str, e);
```
oconnor663: I'm a little skeptical of this change. A common case where you hit an error here is when you do something like `b3sum *`, when the current directory also contains subdirectories. You'll see something like this:

```
$ b3sum *
af1349b9f5f9a1a6a0404dea36dcc9499bcb25c9adc112b7cc9a93cae41f3262  a
b3sum: b: Is a directory (os error 21)
af1349b9f5f9a1a6a0404dea36dcc9499bcb25c9adc112b7cc9a93cae41f3262  c
```

I'd expect all that output to stick together if I e.g. redirected to a file with `b3sum * > out.txt`. But with this change, only the successes would go to the file, and the errors would stay in the terminal. That seems kind of weird to me. What do you think?

oconnor663: Whichever way we go, we should probably add a comment above this line explaining how we feel about this issue.

b3sum/src/main.rs
phayes: `std::io::copy` is unbounded, so we would just copying bytes infinitely since `blake3::OutputReader` never produces an `EOF`.  

I don't see any `std::io::copy_n` or equivalent that is limited to `n` bytes.    This would be a nice edition to the standard library honestly.

b3sum/src/main.rs
phayes: Done.

b3sum/tests/test.rs
phayes: Done.  Nice!

b3sum/tests/test.rs
phayes: Good catch!  

This is now fixed.

b3sum/src/main.rs
phayes: This is my two-cents opinion.  But I'm happy to do whatever you think is best. 

When we have a command-line tool that might "partially fail / partially succeed", as is the case here, we should write only valid expected output to stdout, and leave all the human-readable error messaging to stderr.  This is because we might be piping the output to some other process that will be parsing the output, and partial failure might be perfectly OK (As long as they are reported). 

We then use exit codes (as we are doing), to signal if there were any errors at all. 

Again - this is my two-cents opinion.  Happy to take your direction on whatever you choose.

b3sum/src/main.rs
phayes: done

phayes:
@oconnor663 - I think that's all issues resolved with the exception of your decision on using `println!` vs `eprintln!` for errors.  When you make that decision I'll do a final sweep implementing it and then I think this will be ready.

b3sum/src/main.rs
tom-adsfund: @oconnor663 I think in the case you've mentioned it makes complete sense for failures to go into the error; if someone wanted the errors to go in standard output they could feed the error output into the same stream. Splitting them up like that is a benefit!

b3sum/src/main.rs
phayes: I just learned about `Reader::take()` so this is now fixed.

b3sum/src/main.rs
oconnor663: I was on the fence about this, so as a tiebreaker I looked at what Coreutils does (`md5sum`, `b2sum`, etc). Looks like they all use stderr for the per-file errors. That's good enough for me :)

oconnor663:
LGTM. Did you want to do a final roll-up merge, or shall I merge it as is?

phayes:
Go ahead and merge as is I think.

oconnor663:
Just shipped as part of v0.1.2.

tom-adsfund:
Excellent work, thanks guys!

PR #24, opened 2020-01-14
Adding from_hex and implementing FromStr for Hash

phayes:
This PR adds:

1. `from_hex` for `Hash` as the inverse of `to_hex`, for completeness and convenience. 
2. `FromStr` for `Hash`
3. Adds tests for both `to_hex` and `from_hex`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -177,12 +177,18 @@ impl Hash {
         s
     }
 
-    /// Parse a fixed-length hexidecimal string and return the resulting Hash.
+    /// Parse a hexidecimal string and return the resulting Hash.
     ///
-    /// [`ArrayString`]: https://docs.rs/arrayvec/0.5.1/arrayvec/struct.ArrayString.html
-    pub fn from_hex(hex: ArrayString<[u8; 2 * OUT_LEN]>) -> Result<Self, ParseError> {
+    /// The string must be 64 characters long, producting a 32 byte digest.
+    /// All other string length will return a `ParseError::InvalidLen`.
+    pub fn from_hex(hex: &str) -> Result<Self, ParseError> {
+        let str_bytes = hex.as_bytes();
+        if str_bytes.len() != OUT_LEN * 2 {
+            return Err(ParseError::InvalidLen);
+        }
+
         let mut bytes: [u8; OUT_LEN] = [0; OUT_LEN];
-        for (i, pair) in hex.as_str().as_bytes().chunks(2).enumerate() {
+        for (i, pair) in str_bytes.chunks(2).enumerate() {
             bytes[i] = hex_val(pair[0])? << 4 | hex_val(pair[1])?;
         }
 
@@ -217,9 +223,7 @@ impl core::str::FromStr for Hash {
     type Err = ParseError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let string =
-            ArrayString::<[u8; OUT_LEN * 2]>::from(s).map_err(|_| ParseError::InvalidLen)?;
-        Hash::from_hex(string)
+        Hash::from_hex(s)
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -113,6 +113,16 @@ const KEYED_HASH: u8 = 1 << 4;
 const DERIVE_KEY_CONTEXT: u8 = 1 << 5;
 const DERIVE_KEY_MATERIAL: u8 = 1 << 6;
 
+/// Errors from parsing hex values
+#[derive(Debug)]
+pub enum ParseError {
+    /// Hexadecimal str contains invalid character
+    InvalidChar,
+
+    /// Invalid str length. Only 32 byte digests can be parsed from a 64 char hex encoded str.
+    InvalidLen,
+}
+
 #[inline]
 fn counter_low(counter: u64) -> u32 {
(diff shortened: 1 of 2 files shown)


phayes:
Oops, doesn't looks like I did `no_std` hard enough.  Will fix and update.

phayes:
The only thing I'm really unsure about here is the use of `&static str` as an error-type. I could add a proper error enum for this easily enough.

oconnor663:
Thanks for this PR, and especially for including tests. A couple thoughts:

When I've run into this sort of thing, like in test vectors, I've usually found it easier to go the other way, to convert bytes into hex rather than hex into bytes. The main difference is that you never need to deal with error cases. The downside is that string comparisons aren't constant time, but I haven't found myself needing to do this in anything resembling production code. If going from hex back to `Hash` isn't common, there might not be enough usage to justify maintaining this feature. What do you think? (Note that users can always do this themselves, by parsing hex with e.g. the `hex` crate, and then doing a conversion from bytes.)

If we do want to support this, I imagine it would be more convenient for users to accept a `&str`, rather than an `ArrayString`. Forcing users to do the `ArrayString` conversion themselves make the feature a lot less convenient, especially since it'll usually entail adding a direct dependency on `arrayvec`.

phayes:
@oconnor663, I totally agree about using `&str` - this is now fixed.

oconnor663:
I'm leaning against including this, unless there's a significant use case.

phayes:
Hi @oconnor663,

I would say that including this vs not including this depends on how much you want to recommend hexadecimal as the preferred encoding format for hash digests. 

As for a use-case, this is what is looks like securely comparing two hashes that I have received from elsewhere with and without this PR. 

Without the PR
```rust
use hex;
use blake3;

let hash1 = "7e9e4070a44c39a1b7f836e74383759aadcd341da9a3a71c8ecc3d2630c4f8d1";
let hash2 = "f81f4ad9582dac20141936d2d3930f21a57afd5a04cfed77cf4488e7da6e1196";

let hash1_bytes = hex::decode(hash1)?;
let hash2_bytes = hex::decode(hash2)?;

let hash1_array: [u8; 32];
let hash2_array: [u8; 32];

hash1_array.copy_from_slice(&hash1_bytes[..32]); // might panic
hash2_array.copy_from_slice(&hash2_bytes[..32]);; // might panic

let digest_1: Hash = hash1_array.into();
let digest_2: Hash = hash2_array.into();

if (hash1_digest != hash2_digest) {
    println!("Digests don't match")
}
```

With the PR
```rust
use blake3;

let hash1 = "7e9e4070a44c39a1b7f836e74383759aadcd341da9a3a71c8ecc3d2630c4f8d1";
let hash2 = "f81f4ad9582dac20141936d2d3930f21a57afd5a04cfed77cf4488e7da6e1196";

let digest_1: Hash = hash1.parse()?;
let digest_2: Hash = hash2.parse()?;

if (digest_1 != digest_2) {
    println!("Digests don't match")
}

```

One idea for inclusion might be to hide this behind a `serde` feature flag, and we go ahead and implement `Serialize` and `Deserialize` as well. This would actually be super nice since by default serde will serialize Hash as huge array of integers, which is super sub-optimal.   Again, this comes back to how much you want to recommend hex as the serialization format for `Hash` and how much support you want to give for that recommendation.

oconnor663:
I think if the user is conscientious enough to think about casting into `blake3::Hash` for constant time equality, they could also just as well do this:

```rust
use constant_time_eq::constant_time_eq;

let hash1 = "7e9e4070a44c39a1b7f836e74383759aadcd341da9a3a71c8ecc3d2630c4f8d1";
let hash2 = "f81f4ad9582dac20141936d2d3930f21a57afd5a04cfed77cf4488e7da6e1196";

if !constant_time_eq(hash1.as_bytes(), hash2.as_bytes()) {
    println!("Digests don't match")
}
```

That approach avoids dealing with parse errors, and it probably performs better to boot.

I'm going to close this without merging for now. Parsing hex feels much heavier-weight to me than emitting hex (for example the introduction of the `ParseError` enum), and I'd rather not go there unless there's a strong need.

zaynetro:
My usecase involves serializing hashes and later reading them from disk for comparison. I pretty much prefer parsing them into `Hash` as it gives a clear picture what that field contains.

I agree that this library doesn't need to support reading hashes from hex. Although, it would be really nice to include some kind of documentation on how to do it.

Here is an example that works for me:

```rust
use std::convert::TryInto;
// hex is an external crate

let checksum = "993d1fe78155e0a92bb2e131946eca095bec8196dfd9d81b4b7d919b39b114a6";
let checksum_bytes = hex::decode(checksum).expect("decode hex");
let bytes: [u8; 32] = (&checksum_bytes[..32]).try_into().expect("slice to array");
let hash: Hash = bytes.into();
```

What do you think?

oconnor663:
Good idea. I just added an example to the docs in that commit above.

oconnor663:
I've been thinking more about this over time, and my feelings have changed a bit. (How embarrassing! 😄) I think if we do do this, it might be better to return `Option` rather than trying to provide a meaningful error type. Anyway, if I end up checking something in, I'll make sure to base it on phayes' commits to give credit where credit is due.

oconnor663:
I've pushed the above merge commit and added some follow-on changes.

PR #26, opened 2020-01-15
b3sum: do not mmap files smaller than 16 KiB

### LINKED ISSUE #23: md5sum is slightly faster than b3sum on tiny files
```bash
$ for i in {1..1000};do echo $i > f$i;done
$ hyperfine 'md5sum f{1..1000}' 'b3sum f{1..1000}'
Benchmark #1: md5sum f{1..1000}
  Time (mean ± σ):      11.2 ms ±   0.6 ms    [User: 5.9 ms, System: 5.2 ms]
  Range (min … max):    10.3 ms …  13.5 ms    196 runs
 
Benchmark #2: b3sum f{1..1000}
  Time (mean ± σ):      13.7 ms ±   0.8 ms    [User: 6.2 ms, System: 7.3 ms]
  Range (min … max):    12.8 ms …  20.7 ms    168 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Summary
  'md5sum f{1..1000}' ran
    1.22 ± 0.09 times faster than 'b3sum f{1..1000}'
```

---

I guess the reason is that `read` is more effective than `mmap` on small files.

Strace output (md5sum):
```c
openat(AT_FDCWD, "f998", O_RDONLY)      = 3
fadvise64(3, 0, 0, POSIX_FADV_SEQUENTIAL) = 0
fstat(3, {st_mode=S_IFREG|0644, st_size=4, ...}) = 0
read(3, "998\n", 32768)                 = 4
read(3, "", 28672)                      = 0
lseek(3, 0, SEEK_CUR)                   = 4
close(3)                                = 0
```

Strace output (b3sum):
```c
openat(AT_FDCWD, "f998", O_RDONLY|O_CLOEXEC) = 3
statx(3, "", AT_STATX_SYNC_AS_STAT|AT_EMPTY_PATH, STATX_ALL, {stx_mask=STATX_ALL, stx_attributes=0, stx_mode=S_IFREG|0644, stx_size=4, ...}) = 0
mmap(NULL, 4, PROT_READ, MAP_SHARED, 3, 0) = 0x7fd03e7d2000
munmap(0x7fd03e7d2000, 4)               = 0
close(3)                                = 0
```

---

```
b3sum 0.1.1
md5sum (GNU coreutils) 8.31
```

xzfc:
closes #23

I also have tried disabling mmap at all to emulate heuristic based on the number of files (https://github.com/BLAKE3-team/BLAKE3/issues/23#issuecomment-574703452), but it impacts the Linux kernel source case, so it's not worth it.

# Benchmarks

1000 tiny files:
`for i in {1..1000};do echo $i > f$i;done; hyperfine 'b3sum f{1..1000}'`
```
14.9 ms (this PR)
21.6 ms (prior this PR)
17.7 ms (md5sum)
13.2 ms (do not mmap at all)
```

Linux kernel sources:
`cd linux-5.4.12; time sh -c 'find -type f | xargs b3sum > /dev/null'`
```
3.359 s (this PR)
3.826 s (prior this PR)
3.612 s (md5sum)
4.256 s (do not mmap at all)
```

### DIFF
--- a/b3sum/src/main.rs
+++ b/b3sum/src/main.rs
@@ -78,6 +78,9 @@ fn maybe_memmap_file(file: &File) -> Result<Option<memmap::Mmap>> {
         // Mapping an empty file currently fails.
         // https://github.com/danburkert/memmap-rs/issues/72
         None
+    } else if file_size < 16 * 1024 {
+        // Mapping small files is not worth it.
+        None
     } else {
         // Explicitly set the length of the memory map, so that filesystem
         // changes can't race to violate the invariants we just checked.


phayes:
@xzfc - do we have any idea why 16KiB is the correct number? 

Does 16KiB map to some buffer size in a system-call or something else like that?  Or is it just empirical testing that lead to this number?

xzfc:
@phayes
I've tested this on my machine. I think the exact boundary isn't that important at this scale, the main impact was caused by tiny (less than 32 bytes) files.


```
 8KiB 3.450 s
16KiB 3.312 s
24KiB 3.416 s
32KiB 3.444 s
48KiB 3.564 s
```

oconnor663:
Very simple! Looks great. I'd like to play with it on my machine before merging; hopefully I'll have time today or tomorrow.

oconnor663:
As an aside, if we *really* wanted to have `b3sum` chew through small files, what we'd need to do is hash blocks from separate files in parallel. This is implemented for BLAKE2 e.g. here: https://docs.rs/blake2b_simd/0.5.10/blake2b_simd/many/index.html. It's very high performance, but also kind of complicated. The biggest value add of BLAKE3 is basically that you get the high performance of this parallel mode, but for single files.

If we did that, we'd also want to do a multithreaded directory traversal, like ripgrep does. There'd be some sort of global queue accepting tiny files from all these traversing workers and hashing them in parallel batches. Tuning all of this properly would be...involved :)

oconnor663:
Just shipped as part of v0.1.2.

PR #34, opened 2020-01-19
work in progress to remove arrayvec as a dependency

WildCryptoFox:
[`arrayvec`](https://docs.rs/arrayvec) contains a fair bit of unsafe code ~~and is generally preferred against; in favor of straight up arrays or [`tinyvec`](https://docs.rs/tinyvec)~~. Tinyvec is 100% safe, but requires the containing items to implement `Default` and doesn't yet support `String`s (because `&mut str` isn't awfully useful in Rust without unsafe code).

I'm not sure this BLAKE3 implementation needs either `ArrayVec` implementation; does the state ever need to grow past the allocated array? If not, it can be replaced with just an array. If necessary, `tinyvec::TinyVec` does support switching over to the heap when the collection exceeds the array allocation (with the `alloc` feature).

A couple tests are failing but I cannot pin these down. Any ideas?

**Edit**: I would also suggest against `arrayref` too because `std::convert::TryInto` exists; but as usual this is highly restricted until const-generics land and once they do; [`arrayref` is set to go safe anyway](https://github.com/droundy/arrayref/issues/18) and the unsafe code does look correct.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,6 @@ std = []
 
 [dependencies]
 arrayref = "0.3.5"
-arrayvec = { version = "0.5.1", default-features = false, features = ["array-sizes-33-128"] }
 constant_time_eq = "0.1.5"
 # A performance note for the "rayon" feature: Multi-threading can have
 # significant overhead for small inputs, particularly on x86 where individual
@@ -34,6 +33,11 @@ constant_time_eq = "0.1.5"
 # it for your specific use case.
 rayon = { version = "1.2.1", optional = true }
 cfg-if = "0.1.10"
+
+# replace with crates.io version when available (probably 0.3.0 or 0.2.1)
+[dependencies.tinyvec]
+git = "https://github.com/Lokathor/tinyvec"
+features = ["nightly_const_generics", "alloc"]
 
 [dev-dependencies]
 page_size = "0.4.1"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,7 +59,6 @@ pub mod portable;
 pub mod sse41;
 
 use arrayref::{array_mut_ref, array_ref};
-use arrayvec::{ArrayString, ArrayVec};
 use core::cmp;
 use core::fmt;
 use platform::{Platform, MAX_SIMD_DEGREE, MAX_SIMD_DEGREE_OR_2};
@@ -150,22 +149,6 @@ impl Hash {
     pub fn as_bytes(&self) -> &[u8; OUT_LEN] {
         &self.0
     }
-
-    /// The hexadecimal encoding of the `Hash`. The returned [`ArrayString`] is
-    /// a fixed size and doesn't allocate memory on the heap. Note that
-    /// [`ArrayString`] doesn't provide constant-time equality checking, so if
-    /// you need to compare hashes, prefer the `Hash` type.
-    ///
-    /// [`ArrayString`]: https://docs.rs/arrayvec/0.5.1/arrayvec/struct.ArrayString.html
-    pub fn to_hex(&self) -> ArrayString<[u8; 2 * OUT_LEN]> {
-        let mut s = ArrayString::new();
-        let table = b"0123456789abcdef";
-        for &b in self.0.iter() {
-            s.push(table[(b >> 4) as usize] as char);
-            s.push(table[(b & 0xf) as usize] as char);
-        }
-        s
-    }
 }
 
 impl From<[u8; OUT_LEN]> for Hash {
@@ -202,7 +185,25 @@ impl Eq for Hash {}
 
 impl fmt::Debug for Hash {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "Hash({})", self.to_hex())
+        write!(f, "Hash({:x})", self)
+    }
+}
+
+impl fmt::LowerHex for Hash {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        for b in self.as_bytes() {
+            write!(f, "{:02x}", b)?;
+        }
+        Ok(())
+    }
+}
+
+impl fmt::UpperHex for Hash {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        for b in self.as_bytes() {
+            write!(f, "{:02X}", b)?;
+        }
+        Ok(())
     }
 }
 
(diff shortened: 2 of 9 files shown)


src/lib.rs
```diff
@@ -150,22 +149,6 @@ impl Hash {
...
-    /// [`ArrayString`]: https://docs.rs/arrayvec/0.5.1/arrayvec/struct.ArrayString.html
-    pub fn to_hex(&self) -> ArrayString<[u8; 2 * OUT_LEN]> {
-        let mut s = ArrayString::new();
-        let table = b"0123456789abcdef";
-        for &b in self.0.iter() {
-            s.push(table[(b >> 4) as usize] as char);
-            s.push(table[(b & 0xf) as usize] as char);
-        }
-        s
-    }
```
WildCryptoFox: A slight breaking change. See the `Display` implementation below for the standard approach to write out bytes as hex in Rust. I can split into its own PR if preferred.

src/lib.rs
```diff
@@ -447,9 +439,11 @@ fn compress_chunks_parallel(
     debug_assert!(input.len() <= MAX_SIMD_DEGREE * CHUNK_LEN);
 
     let mut chunks_exact = input.chunks_exact(CHUNK_LEN);
-    let mut chunks_array = ArrayVec::<[&[u8; CHUNK_LEN]; MAX_SIMD_DEGREE]>::new();
-    for chunk in &mut chunks_exact {
-        chunks_array.push(array_ref!(chunk, 0, CHUNK_LEN));
+    let mut chunks_array = [&[0u8; CHUNK_LEN]; MAX_SIMD_DEGREE];
```
WildCryptoFox: `CHUNK_LEN = 64` but `[T; 64]: Default` is not true until const-generics land; so this has been expanded to just an array and associated length.

Cargo.toml
```diff
@@ -35,6 +34,11 @@ constant_time_eq = "0.1.5"
 rayon = { version = "1.2.1", optional = true }
 cfg-if = "0.1.10"
 
+# replace with crates.io version when available (probably 0.3.0 or 0.2.1)
+[dependencies.tinyvec]
+git = "https://github.com/Lokathor/tinyvec"
+features = ["nightly_const_generics", "alloc"]
```
WildCryptoFox: nightly const-generics feature necessary for its `Array` impls on larger sized collections, notably `MAX_DEPTH`.

WildCryptoFox: This can be easily fixed without nightly by asking @Lokathor to include 54 too. I did get the `const-generics` added for this.

Lokathor:
please be advised that i was going to put out another release soon with some more important bug fixes.

I'll try to get 0.3 tomorrow!

src/lib.rs
```diff
@@ -497,9 +490,11 @@ fn compress_parents_parallel(
...
-    for parent in &mut parents_exact {
-        parents_array.push(array_ref!(parent, 0, BLOCK_LEN));
+    let mut parents_array = [&[0u8; BLOCK_LEN]; MAX_SIMD_DEGREE_OR_2];
+    let mut parents_so_far = 0;
+    for (a, b) in parents_array.iter_mut().zip(&mut parents_exact) {
+        *a = array_ref!(b, 0, BLOCK_LEN);
+        parents_so_far += 1;
     }
     platform.hash_many(
         &parents_array,
```
sneves: `parents_array` is not guaranteed to be `MAX_SIMD_DEGREE_OR_2` here; it can range anywhere from 2 to `MAX_SIMD_DEGREE`, so you need to pass the actual size somehow.

oconnor663: This is definitely the cause of at least some of the test failures.

oconnor663:
Disorganized thoughts:

- This seems like a lot of churn for a change with theoretical benefits. If there were a significant speed-up in build times, or a known problem with `arrayvec`, that would be different.
- `arrayvec` is mature and widely used. Is there a source for "generally preferred against"?
- Soundness is obviously a top priority for this crate, as it is for most crates. However, because this is cryptography, any sort of logic bug could also be a serious security issue. For that reason, "using less unsafe code in dependencies" needs to be weighed against "using mature and well-tested dependencies". I'm sure `tinyvec` is going to be an excellent crate as it grows, but it looks like it's only a couple weeks old.
- `blake3::Hash::to_hex`, based on `arrayvec::ArrayString`, is a very convenient feature, and it's important that it doesn't allocate memory. Replacing it with `Display` creates two problems: Even though `Display` theoretically supports `no_std` callers, the only convenient interface I'm aware of is `to_string`, which allocates a `String` on the heap. And it's not clear to me that "to string" should necessarily imply "to hex". That could get confusing if we ever add e.g. Base64 serialization. Note that many fundamental types like `&[u8]` don't implement `Display`.

It looks like you're putting a lot of work into this, and I will be happy to review it more carefully when you think it's ready. But I want to be up front that my leaning right now is not to merge.

Lokathor:
I agree it's probably soon.

I also am not totally convinced that `tinyvec` is without any performance cost.

WildCryptoFox:
This PR is marked as a draft and is not intended to be merged immediately. I just hit a roadblock and couldn't debug it at the time. Thanks for the insight @sneves.

"generally preferred against", I was referring to the unsafety, not specifically `arrayvec`. `tinyvec` may be new but I've consistently seen a move towards it. Maybe this view is biased or too early. I can see how poorly I wrote that... I'll just strike it out. :)

I can say I prefer `tinyvec`, I just wish const-generics would land so `Default` would cover all the array sizes.

src/lib.rs
WildCryptoFox: @sneves Okay.. That's odd. I did try adding `[..parents_so_far]` after `&parents_array` and that didn't work... Now it does. Maybe a different write was skipped by accident somewhere else. *shrugs*. Thanks.

WildCryptoFox:
@oconnor663 I can see the appeal for `arrayvec::ArrayString` but if someone is writing to a string; they'll have their own buffer to write into. Let that be the stdout buffer, or a `String` or another `ArrayString`. `Display` is more flexible and does not force an allocation either. The `Deref<&str>` of `ArrayString` doesn't seem any more useful than `Display` besides caching if it is used many times.

```rust
write!(out, "hash: {}", my_hash);
```

cormacrelf:
@oconnor663 

> Even though Display theoretically supports no_std callers, the only convenient interface I'm aware of is to_string, which allocates a String on the heap.

`to_string` is not the only one. `core::fmt::Write` is implemented by `ArrayString`. So you can write to a fixed-size string directly,  in `no_std` environments.

```rust
use arrayvec::ArrayString; // 0.5.1
use core::fmt::Write;

fn main() {
    let x = 0x7b;
    let mut buf = ArrayString::<[u8; 20]>::new();
    write!(&mut buf, "{:02x}", x).expect("Can't write");
    assert_eq!(&buf, "7b");
}
```

This

> And it's not clear to me that "to string" should necessarily imply "to hex". That could get confusing if we ever add e.g. Base64 serialization.

...is solved by implementing `core::fmt::LowerHex` and `UpperHex` instead of `Display`. I wouldn't recommend actually adding base64, unless you enjoy mundane issues about RFC4648 and url-safe variants, but this prevents confusion at least. Probably needs an example in the docs for discoverability, but I think it's a better choice than Display, if only because people using it need either upper or lower hex output, you can't tell by `"{}"` which it's going to be, and some people's assumptions will be wrong, especially if they use Windows where most hex strings are uppercase.

You can still get heap strings pretty quickly with `format!("{:x}", hash)`.

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=3865d67984559b981090e8b519b17ddc

WildCryptoFox:
Speaking of base64 and hex. @tarcieri has [constant time implementations](https://docs.rs/subtle-encoding/0.5.0/subtle_encoding/).

WildCryptoFox:
@cormacrelf `s/Display/{Lower,Upper}Hex/` applied.

oconnor663:
Going to close this for now, as I'm comfortable with `ArrayVec` and `to_hex` as they are.

Issue #32, opened 2020-01-18
b3sum incredibly slow


tERyceNzAchE:
b3sum runs incredibly slow.

PC:
Windows 10 Pro
i7-6700
32GB RAM

b3sum built using: `cargo install b3sum` (v0.1.2)

Test File: 2.88GB

Results:
```
md5sum: 10 seconds
b2sum: 5 seconds
b3sum: 4 Minutes 37 Seconds
```

oconnor663:
Out of curiosity, what happens if you hash via stdin like this: `b3sum < big_file` (does that work on Windows)

tERyceNzAchE:
That took it from minutes to seconds!

oconnor663:
Cool. "Thrashing the disk" is my current theory for your slow numbers. The next thing I'd be curious to try is a file small enough to fit in memory. That should work better with all the threads, because you won't actually be hitting the disk do much. Also, do you happen to have a spinning drive?

tERyceNzAchE:
Yes, this is from a spinning drive.

oconnor663:
"Don't bother with multi-threading when you're on a spinning drive and the file is too small to fit in cache" would be a decent performance heuristic, but I don't know of any simple or reliable way to detect either of those two conditions...

oconnor663:
You know, I wonder if the performance problem here is *audible* :)

tERyceNzAchE:
> I wonder if the performance problem here is audible :)

It does seem to be.

tERyceNzAchE:
> "Don't bother with multi-threading when you're on a spinning drive and the file is too small to fit in cache" would be a decent performance heuristic, but I don't know of any simple or reliable way to detect either of those two conditions...

This isn't a small file.  Perhaps a command line flag to disable multi-threading?

oconnor663:
Oops, meant to say "too large" above. Some flag like that might be useful for benchmarking and testing (situations like this), and we could totally add it. That said, I don't expect that most users facing this sort of issue will be able to discover that the `--single-threaded` flag (or whatever we call it) fixes things.

xzfc:
This isn't documented in b3sum itself, but you can disable multi-threading by setting `RAYON_NUM_THREADS=1` environment variable.

tERyceNzAchE:
> This isn't documented in b3sum itself, but you can disable multi-threading by setting RAYON_NUM_THREADS=1 environment variable.

[...]

PR #36, opened 2020-01-20
b3sum: add no-mmap option

cesarb:
Using mmap is not always the best option. For instance, if the file is
truncated while being read, b3sum will receive a SIGBUS and abort.

Follow ripgrep's lead and add a --no-mmap option to disable mmap. This
can also help benchmark the mmap versus the read path, and help debug
performance issues potentially caused by mmap access patterns (like
issue #32).


oconnor663:
Excellent idea. Thrilled by the test case :)

PR #37, opened 2020-01-20
Use large buffer for no-mmap

cesarb:
Instead of the default 8192-byte buffer, use a large buffer when not
using mmap. Also add a hidden command line option to vary the buffer
size for benchmarking.

The default value chosen is 256 KiB per logical core. This size should
be smaller than the L3 cache for most common desktop processors.

This pull request depends on https://github.com/BLAKE3-team/BLAKE3/pull/36 due to conflicts.

Unfortunately, this is a failed experiment. Testing with a 3 GiB zero-filled file on tmpfs, the fastest buffer size is 8192 bytes, exactly the size `std::io::copy` uses; even using mmap is slower. I'm opening this pull request in case you want to try for yourself, but feel free to close it without merging.

### DIFF
--- a/b3sum/Cargo.toml
+++ b/b3sum/Cargo.toml
@@ -20,6 +20,7 @@ blake3 = { version = "0.1", path = ".." }
 clap = { version = "2.33.0", default-features = false }
 hex = "0.4.0"
 memmap = { version = "0.7.0", optional = true }
+num_cpus = "1.0"
 
 [dev-dependencies]
 assert_cmd = "0.12.0"
--- a/b3sum/src/main.rs
+++ b/b3sum/src/main.rs
@@ -9,6 +9,7 @@ const FILE_ARG: &str = "file";
 const LENGTH_ARG: &str = "length";
 const KEYED_ARG: &str = "keyed";
 const DERIVE_KEY_ARG: &str = "derive-key";
+const NO_MMAP_ARG: &str = "no-mmap";
 const NO_NAMES_ARG: &str = "no-names";
 const RAW_ARG: &str = "raw";
 
@@ -37,6 +38,11 @@ fn clap_parse_argv() -> clap::ArgMatches<'static> {
                 .takes_value(true)
                 .value_name("CONTEXT")
                 .help("Uses the key derivation mode, with the input as key material"),
+        )
+        .arg(
+            Arg::with_name(NO_MMAP_ARG)
+                .long(NO_MMAP_ARG)
+                .help("Never use memory maps"),
         )
         .arg(
             Arg::with_name(NO_NAMES_ARG)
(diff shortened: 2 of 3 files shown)


oconnor663:
You will definitely see performance gains if you try this on an AVX-512-supporting machine, because the 8192-byte buffer used in `std::io::copy` is only large enough to drive AVX2.

I've tried a double-buffering approach here: https://github.com/BLAKE3-team/BLAKE3/commit/e2b200f1daccecf059bd206da6f667f8a4e40382. Performance-wise, I think that implementation got pretty close to the memory-mapped version, so that was a good sign. But for a fallback code path, I wasn't happy with how complicated it felt.

oconnor663:
Which is to say, I think we should definitely do something like this, and I'm still bouncing the idea around in my head to try to figure out the cleanest way to do it.

cesarb:
The biggest culprit of the slowdown with large buffers seems to be `rayon`. Telling it to use just one thread and moving the thread into the `rayon` thread pool with `ThreadPool::install()` are not enough, you have to compile without `rayon` support. After that, the optimal buffer size at least on my machine seems to be around 128 KiB, which is half the size of the L2 cache.

oconnor663:
Oh right, there are multiple factors here. The implementation uses `rayon::join` as early as possible, which means as soon as the available SIMD lanes are occupied. For typical AVX2 machines, that's any input above 8 KiB. For an AVX-512 machine, that's any input above 16 KiB. This is usually bad for performance below about 128 KiB, but I've left it in because there are also systems where it's good (slower ARM devices), and I've found it hard to define a simple rule for what the threshold should be. But maybe you could suggest such a rule?

So getting good multi-threaded performance out of the incremental mode requires two things:

1) A very large buffer, at least 1 MiB.
2) A double-buffering strategy that avoids blocking all the Rayon work while the next buffer is being filled.

cesarb:
Another thing which can help getting good performance of incremental mode is to run the whole update loop within a `ThreadPool::install()` call (that is, within a `rayon` thread), since otherwise every top-level `rayon::join` call (that is, every call to `Hasher::update`) sends both the left and right operations to the thread pool, instead of running the left operation in the caller thread. That makes more of a difference with smaller buffer sizes (since only the top-level `rayon::join` call has this issue). As a bonus, it would allow easily configuring the parallelism through a command line parameter.

Unfortunately, I don't have any good idea on how to choose the split threshold. It's probably around the point where the `rayon` overhead is the same as the time to complete another unit of work. Since an out-of-order PC with wide vectors can do an unit of work so fast, it can cost more time to put it in the `rayon` queues than to do it directly in the same thread; while on an in-order low-speed core, the same unit of work can take long enough that it would take longer to run it in the same thread than the time it takes to start it on an idle thread. That is, it depends both on the speed of completing an unit of work in a single thread, and on the speed of sending that unit of work to another thread.

oconnor663:
Going to close this for now, but we should revisit at some point with a fancier background thread strategy.

PR #38, opened 2020-01-22
Fix popcnt implementation when building for 32-bit

erijo:
When building with `-m32`, gcc would previously complain about:
```
blake3_impl.h:49:19: warning: implicit declaration of function ‘_mm_popcnt_u64’; did you mean ‘_mm_popcnt_u32’? [-Wimplicit-function-declaration]
   49 |   return (uint8_t)_mm_popcnt_u64(x);
      |                   ^~~~~~~~~~~~~~
      |                   _mm_popcnt_u32
```

Fix this by using `_mm_popcnt_u32` in that case.

### DIFF
--- a/c/blake3_impl.h
+++ b/c/blake3_impl.h
@@ -70,10 +70,20 @@ static const uint8_t MSG_SCHEDULE[7][16] = {
     {11, 15, 5, 0, 1, 9, 8, 6, 14, 10, 2, 12, 3, 4, 7, 13},
 };
 
+INLINE uint32_t counter_low(uint64_t counter) { return (uint32_t)counter; }
+
+INLINE uint32_t counter_high(uint64_t counter) {
+  return (uint32_t)(counter >> 32);
+}
+
 // Count the number of 1 bits.
 INLINE uint8_t popcnt(uint64_t x) {
 #if __POPCNT__
+#ifdef __x86_64__
   return (uint8_t)_mm_popcnt_u64(x);
+#else
+  return (uint8_t)(_mm_popcnt_u32(counter_low(x)) + _mm_popcnt_u32(counter_high(x)));
+#endif
 #else
   uint8_t count = 0;
   while (x > 0) {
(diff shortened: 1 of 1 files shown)


c/blake3_impl.h
```diff
@@ -43,10 +43,20 @@ static const uint8_t MSG_SCHEDULE[7][16] = {
...
+INLINE uint32_t counter_low(uint64_t counter) { return (uint32_t)counter; }
+
+INLINE uint32_t counter_high(uint64_t counter) {
+  return (uint32_t)(counter >> 32);
+}
+
 // Count the number of 1 bits.
 INLINE uint8_t popcnt(uint64_t x) {
 #if __POPCNT__
+#ifdef __x86_64__
```
oconnor663: Do we need to check an extra variable for MSVC support here? In `blake3_dispatch.c` we also check `_M_X64`. Then again, maybe we have the same problem with the `__POPCNT__` variable.

Makes me wonder if this optimization is even worth it. The portable version is almost certainly not a hot spot.

erijo: I did a quick test using a [msvc online compiler](https://rextester.com/l/cpp_online_compiler_visual) that I found and it looks like `__POPCNT__` isn't defined (or I just don't know the correct compiler argument to use).

erijo:
In the [msvc documentation](https://docs.microsoft.com/en-us/cpp/intrinsics/popcnt16-popcnt-popcnt64?view=vs-2019) for the popcnt intrinsics they state:
> In 32-bit mode, there are no 64-bit general-purpose registers, so 64-bit popcnt isn't supported.

Both GCC and Clang have the definition of `_mm_popcnt_u64` guarded by `#ifdef __x86_64__`.

c/blake3_impl.h
sneves: MSVC has no way to detect popcnt at compile-time.

oconnor663:
We just landed Samuel's commit 37ea737c, which rewrites `popcnt`. Does the current version work on your platform?

erijo:
Yes, that seems to work (tested with `CFLAGS += -m32` and `make test`).

PR #40, opened 2020-01-23
Add extern "C" to blake3.h

erijo:
So that the header can be included in C++-programs without getting linker errors.


oconnor663:
I don't know the best practices here, so I'll defer to @sneves.

sneves:
Yes, this is necessary to work in C++ due to name mangling.

Issue #33, opened 2020-01-18
Enhancement: b3sum add -c --check flag for verification


tERyceNzAchE:
Adding -c, --check would make b3sum a drop in replacement for b2sum in some scripts.

oconnor663:
Agreed, this would be a good feature. It's not high on my personal priority list, because I don't use `--check` myself, but I'd be happy to review a PR.

phayes:
I would be happy to create a PR for this. 

But first I think I would like @oconnor663's decision on #24 since that will determine if the hex support is in `b3sum` or in `blake3`

oconnor663:
Apologies for dragging my feet on these. I'm hoping I'll get to them by the end of the week.

phayes:
@oconnor663 , no worries!  It's your project - take your time.  :)

oconnor663:
I decided against #24, so let's go with putting the hex parsing in `b3sum`. (Or for what it's worth, we could do constant-time hex comparisons instead of parsing.)

phayes:
OK great.   I'll have this done in the next day or two.

oconnor663:
Exciting! I know the https://github.com/jtdowney/b2sum-rs project has a `--check` implementation, in case that helps. I'm not sure whether everyone does this the same way, or whether we're going to have to make choices.

phayes:
A work-in-progress is here: #44

llowrey:
Is there interest in completing this feature? It looks like there are a few unresolved issues with PR #44 and those issues appear to be more policy (requirements) related than technical/implementation/defects.

I am a heavy user of "b2sum --check" and would like to use b3sum, for performance reasons. In one trivial use case I am able to check a dir of files in 42s with b3sum that takes 5m37s with b2sum. That's a big difference but atm I have to do the check out-of-band which is neither convenient or reliable (in my case).

I'd like to help get this feature released but not sure how to contribute.

oconnor663:
I'm definitely interested in the abstract. It just hasn't been a priority of mine. There were some policy questions that came up in that PR, which I think it might be a good idea to for us to try to answer before we write much more code. Let me see if I can list them here:

1. **Weird filepath support.** Can filepaths contain e.g. newlines? Can they contain invalid UTF-8? "Do whatever Coreutils does here" could be a reasonable answer, but Coreutils actually does some very non-trivial things, and I'd want to make sure we really understand what those are.
[...]

PR #44, opened 2020-01-26
Adding --check to b3sum

phayes:
This is for adding `--check` to the `b3sum` utility.

### DIFF
--- a/b3sum/src/main.rs
+++ b/b3sum/src/main.rs
@@ -1,4 +1,4 @@
-use anyhow::{bail, Context, Result};
+use anyhow::{bail, Context, Error, Result};
 use clap::{App, Arg};
 use std::cmp;
 use std::convert::TryInto;
@@ -186,12 +186,15 @@ fn read_key_from_stdin() -> Result<[u8; blake3::KEY_LEN]> {
 
 fn check_file(input: impl Read, mmap_disabled: bool) -> Result<u64> {
     let hasher = blake3::Hasher::new();
-    let reader = std::io::BufReader::new(input);
     let mut num_failed = 0;
-    for line in reader.lines() {
-        let line: String = line?;
+
+    let reader = std::io::BufReader::new(input);
+    for (line_num, line) in reader.lines().enumerate() {
+        // Read the line and trim it
+        let line: String = line.with_context(|| format!("line {}", line_num + 1))?;
         let line = line.trim();
 
+        // Skip the line if it's a comment or empty
         if line.starts_with('#') {
             continue;
         }
@@ -199,21 +202,31 @@ fn check_file(input: impl Read, mmap_disabled: bool) -> Result<u64> {
             continue;
         }
 
+        // Split line by whitespace
         let mut line = line.split_whitespace();
 
-        let hash: &str = line.next().unwrap(); // TODO
-        let hash = hex::decode(hash)?;
+        // Get the hash value and parse the hex value
+        let hash: &str = line.next().expect("Unexpected missing value");
+        let hash = hex::decode(hash)
+            .with_context(|| format!("error decoding hex on line {}", line_num + 1))?;
 
-        let filename: &str = line.next().unwrap(); // TODO
+        // Get the filename
+        let filename: &str = line
+            .next()
+            .ok_or_else(|| Error::msg(format!("missing filename on line {}", line_num + 1)))?;
         let os_filename: std::ffi::OsString = filename.into();
 
+        // Hash the file
         let hasher = hasher.clone();
-        let check_reader = hash_file(&hasher, &os_filename, mmap_disabled)?;
         let mut check_hash = Vec::<u8>::with_capacity(hash.len());
+        let check_reader = hash_file(&hasher, &os_filename, mmap_disabled)
+            .with_context(|| format!("filename: {}", filename))?;
         check_reader
             .take(hash.len() as u64)
             .read_to_end(&mut check_hash)?;
 
+        // Check if hashes match.
+        // TODO: Should this be constant time?
         if check_hash == hash {
             println!("{}: OK", filename);
         } else {
@@ -230,8 +243,10 @@ fn check_files(args: clap::ArgMatches) -> Result<u64> {
     let mmap_disabled = args.is_present(NO_MMAP_ARG);
     if let Some(files) = args.values_of_os(FILE_ARG) {
         for filepath in files {
-            let file = File::open(filepath)?;
-            num_failed += check_file(file, mmap_disabled)?;
+            let file = File::open(filepath)
+                .with_context(|| format!("checking {}: ", filepath.to_string_lossy()))?;
+            num_failed += check_file(file, mmap_disabled)
+                .with_context(|| format!("checking {}: ", filepath.to_string_lossy()))?;
         }
     } else {
         let stdin = std::io::stdin();
--- a/b3sum/src/main.rs
+++ b/b3sum/src/main.rs
@@ -12,6 +12,7 @@ const DERIVE_KEY_ARG: &str = "derive-key";
 const NO_MMAP_ARG: &str = "no-mmap";
 const NO_NAMES_ARG: &str = "no-names";
 const RAW_ARG: &str = "raw";
+const CHECK_ARG: &str = "check";
 
 fn clap_parse_argv() -> clap::ArgMatches<'static> {
     App::new("b3sum")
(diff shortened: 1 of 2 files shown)


phayes:
OK I think this is ready for review.

b3sum/src/main.rs
```diff
@@ -177,8 +184,94 @@ fn read_key_from_stdin() -> Result<[u8; blake3::KEY_LEN]> {
...
+        let hasher = hasher.clone();
+        let mut check_hash = Vec::<u8>::with_capacity(hash.len());
+        let check_reader = hash_file(&hasher, &os_filename, mmap_disabled)
+            .with_context(|| format!("filename: {}", filename))?;
+        check_reader
+            .take(hash.len() as u64)
+            .read_to_end(&mut check_hash)?;
+
+        // Check if hashes match.
+        // TODO: Should this be constant time?
```
phayes: @oconnor663 - I'd like your opinion of if we should be doing a constant-time check for equality here.

b3sum/src/main.rs
```diff
@@ -177,8 +184,94 @@ fn read_key_from_stdin() -> Result<[u8; blake3::KEY_LEN]> {
...
+        if check_hash == hash {
```
phayes: @oconnor663 , I'd like your opinion of if we should be doing a constant-time check for equality here.

oconnor663: I don't see any reason not to, and it might matter in some very obscure scenario. Let's stick to constant-time. We could even do a constant-time comparison on the hex bytes directly, and save ourselves the trouble of parsing them.

b3sum/src/main.rs
```diff
@@ -177,8 +184,94 @@ fn read_key_from_stdin() -> Result<[u8; blake3::KEY_LEN]> {
...
+    let hasher = blake3::Hasher::new();
+    let mut num_failed = 0;
+
+    let reader = std::io::BufReader::new(input);
+    for (line_num, line) in reader.lines().enumerate() {
+        // Read the line and trim it
+        let line: String = line.with_context(|| format!("line {}", line_num + 1))?;
+        let line = line.trim();
+
+        // Skip the line if it's a comment or empty
```
oconnor663: Do these conditions come up in practice? What does coreutils do with these cases?

b3sum/src/main.rs
```diff
@@ -177,8 +184,94 @@ fn read_key_from_stdin() -> Result<[u8; blake3::KEY_LEN]> {
...
+        // Get the hash value and parse the hex value
+        let hash: &str = line.next().expect("Unexpected missing value");
+        let hash = hex::decode(hash)
+            .with_context(|| format!("error decoding hex on line {}", line_num + 1))?;
+
+        // Get the filename
+        let filename: &str = line
+            .next()
+            .ok_or_else(|| Error::msg(format!("missing filename on line {}", line_num + 1)))?;
+        let os_filename: std::ffi::OsString = filename.into();
```
oconnor663: This raises an interesting point. Should we try to handle really weird filenames and escaping? It looks like `md5sum * | md5sum --check` can handle some very nasty stuff, including filenames with newlines or invalid unicode.

b3sum/src/main.rs
```diff
@@ -177,8 +184,94 @@ fn read_key_from_stdin() -> Result<[u8; blake3::KEY_LEN]> {
...
+
+        // Hash the file
+        let hasher = hasher.clone();
+        let mut check_hash = Vec::<u8>::with_capacity(hash.len());
+        let check_reader = hash_file(&hasher, &os_filename, mmap_disabled)
+            .with_context(|| format!("filename: {}", filename))?;
+        check_reader
+            .take(hash.len() as u64)
+            .read_to_end(&mut check_hash)?;
```
oconnor663: I'm a little skeptical of deriving the length from the input hashes. Say for example that the input hashes are just 1 byte long. That could lead to a situation where you think you're checking file integrity, but you're really not checking much. This gets more problematic if it's possible for some attacker to _control_ the length of the hashes you check.

On the flip side, checking hashes longer than 32 bytes doesn't add any security, but people might get confused and think that it does.

Maybe just have `--check` always assume the default length? Maybe even conflict with `--length`? (Incidentally it should probably also conflict with `--keyed` and `--derive-key`.)

b3sum/src/main.rs
```diff
@@ -177,8 +184,94 @@ fn read_key_from_stdin() -> Result<[u8; blake3::KEY_LEN]> {
     }
 }
 
+fn check_file(input: impl Read, mmap_disabled: bool) -> Result<u64> {
```
oconnor663: It's a little confusing that you've got `check_files` and `check_file`, but it's the latter that loops over all the filenames?

b3sum/tests/test.rs
```diff
@@ -134,3 +134,16 @@ fn test_raw_with_multi_files_is_an_error() {
...
+    let in_file = tempfile::NamedTempFile::new().unwrap();
+    in_file.as_file().write_all(b"foo").unwrap();
+    in_file.as_file().flush().unwrap();
+
+    // Check the resuts by doing a round-trip via stdin / stdout
+    let result = cmd!(b3sum_exe(), in_file.path())
+        .pipe(cmd!(b3sum_exe(), "--check"))
+        .run();
+    assert!(result.is_ok());
+}
```
oconnor663: We'll definitely want to test multiple inputs. Also whatever we do with weird whitespace and invalid unicode, we should try to test here.

b3sum/src/main.rs
sneves: You would have to lowercase both strings in that case, which is about the same trouble as parsing the hex string into bytes and comparing there.

b3sum/src/main.rs
phayes: I was trying to ensure that all output from `b3sum` is also valid input for `b3sum --check`.    So that if you do `b3sum * --length=128` and pass that back into `b3sum --check` it should work. 

I would recommend two things:

1. Throw a warning if the hash lengths are under some threshold at which we consider the security compromised.  We should do this for both checking and generating the checksum file. 

2. Add a "strict" mode where warnings are upgraded to errors. 

This has some precedence.  `b2sum-rs` has the following options that we could emulate:
```
The following five options are useful only when verifying checksums:
      --ignore-missing  don't fail or report status for missing files
      --quiet           don't print OK for each successfully verified file
      --status         don't output anything, status code shows success
      --strict          exit non-zero for improperly formatted checksum lines
``` 

However, if we do decide to error on checking non-32-byte checksums, then we should make sure we at least generate a warning when generating them.

b3sum/src/main.rs
phayes: I'm a big fan of drawing the line at non-unicode filenames.     Non-unicode filenames need to be relegated to the dustbin of history  :)  

But if you really think we need to support non-unicode filenames, then I can work on adding support. 

I agree that we should add tests for all sorts of pathological (unicode) filename cases though.

b3sum/src/main.rs
oconnor663: Are we allowed to assume that the output we're parsing came from `b3sum`? If so, maybe we don't have to support uppercase hex if we don't emit any? But anyway, parsing hex to bytes is also fine with me.

b3sum/src/main.rs
phayes: I was following along with the behavior here: 

https://github.com/jtdowney/b2sum-rs/blob/master/src/main.rs#L127-L129

phayes: `shasum` (on my mac at least) allows one trailing empty line, but errors on more than one consecutive empty line.

b3sum/src/main.rs
phayes: Correct.  I will rename things to make it clearer.

b3sum/src/main.rs
oconnor663: On the one hand, it might not be worth the effort. On the other hand, if we treat everything as raw bytes, maybe we get this "support" for free? (But then again, Windows paths aren't amenable to raw bytes, are they? Not making Window support an annoyance might be a good reason to stick with Unicode/UTF-8.)

phayes: Also, deciding what is whitespace and a line-end when interpreting raw bytes is a PITA.    I think I would want to bring in `regex::bytes` as a dependency to handle this if we decide to support non-unicode.

b3sum/src/main.rs
oconnor663: If we can simplify things by demanding that the input format is exactly the same as the tool output, I'm all for simplifying.

b3sum/src/main.rs
oconnor663: I think if we have `check` throw an error saying "this only works with 32-byte hashes", we'll get 90% of the value with 10% of the effort. Trying to determine "security compromised" in general is too vague, and it depends on the application. (Collision-resistant checksums need to be 256 bits, but MACs can usually get away with 128, for example.)

oconnor663:
I've pushed a different `--check` implementation to master, and documented it at https://github.com/BLAKE3-team/BLAKE3/blob/master/b3sum/what_does_check_do.md. If you get a chance, and let me know what you think at https://github.com/BLAKE3-team/BLAKE3/issues/33?

Issue #45, opened 2020-01-27
[C/main.c] function memset used before return of malloc was done.


TheVice:
Hello.

In **main.c** _memset_ was call on memory before this block was checked on NULL after it get from _malloc_ function. One of solution placed it after this check was done.

Next [pull request](https://github.com/BLAKE3-team/BLAKE3/pull/46) can be applied to do this.

Thank you.

P.S.
Also **out_len** can be check before _malloc_, but that is different story.

oconnor663:
> Also out_len can be check before malloc, but that is different story.

`main.c` is only used for testing, so performance doesn't particularly matter here.

TheVice:
By the way, creating of 'out' [can be moved](https://github.com/BLAKE3-team/BLAKE3/pull/49) outside of 'do' loop.

Also we can discuss about using preferred size of buffer base on 64 and in this case internal buffer 'wide_buf' [can be removed](https://github.com/BLAKE3-team/BLAKE3/pull/50). This request marked as WIP according to not pass checking at moment when this comment was written.

PR #46, opened 2020-01-27
[memset] placed function after checking of memory was done.

TheVice:
Hello.

This is a one of solution that resolve [this issue](https://github.com/BLAKE3-team/BLAKE3/issues/45).

Thank you.


oconnor663:
Good catch. I think we can get rid of this `memset` entirely, since `blake3_hasher_finalize` is going to completely overwrite the same bytes.

TheVice:
Agree, than remove call of _memset_ function.

Issue #47, opened 2020-01-27
[MSVC] support of Microsoft C compiler is missed.


TheVice:
Hello.

There is a series of issues at source files that prevent to compile project at Microsoft C compiler:
- including of POSIX header;
- applying of unary minus operator to the unsigned value;
- including of header that not applying to the all target platforms (ARM/x86/...).

One of vision how to resolve that all can be found in [this request](https://github.com/BLAKE3-team/BLAKE3/pull/48).

Thank you.

P.S.
Please note, here we talking about compile issue. Linking at MSVC environment have same issue like at gcc that already discussed [here](https://github.com/BLAKE3-team/BLAKE3/issues/16).

oconnor663:
That PR has been merged, so I'm going to go ahead and close this issue. We can reopen this one or another one if more bugs crop up.

Issue #16, opened 2020-01-11
non-test C implementation fails to build


adamdmoss:
gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0

```
% make
gcc -O3 -Wall -Wextra -std=c11 -pedantic -c blake3_sse41.c -o blake3_sse41.o -msse4.1 -D BLAKE3_USE_SSE41
gcc -O3 -Wall -Wextra -std=c11 -pedantic -c blake3_avx2.c -o blake3_avx2.o -mavx2 -D BLAKE3_USE_SSE41 -D BLAKE3_USE_AVX2
gcc -O3 -Wall -Wextra -std=c11 -pedantic -c blake3_avx512.c -o blake3_avx512.o -mavx512f -mavx512vl -D BLAKE3_USE_SSE41 -D BLAKE3_USE_AVX2 -D BLAKE3_USE_AVX512
gcc -O3 -Wall -Wextra -std=c11 -pedantic blake3.c blake3_dispatch.c blake3_portable.c main.c blake3_sse41.o blake3_avx2.o blake3_avx512.o -o blake3
/tmp/ccJZjf2t.o: In function `main':
main.c:(.text.startup+0x1f9): undefined reference to `get_cpu_features'
main.c:(.text.startup+0x245): undefined reference to `g_cpu_features'
collect2: error: ld returned 1 exit status
Makefile:6: recipe for target 'all' failed
make: *** [all] Error 1

```

Similar story with clang version 8.0.0-3~ubuntu18.04.2 (tags/RELEASE_800/final)

dd4rk:
Simply `make` doesn't generate `-DBLAKE3_TESTING`, that causes `enum cpu_feature g_cpu_features` to be static.
Use `make test` or write your own test code instead.

oconnor663:
Yes originally the bare `make` command would compile a working executable (pretty bare bones, mostly just for playing around). As part of adding dynamic feature detection, that build target broke. We need to fix it, of course, but we also need to substantially clean up the docs and CLI interface for that target. It's not yet clear exactly what we want it to be.

oconnor663:
I'm going to close this, as `main.c` is only intended for testing at this point. The new `c/README.md` file has new instructions for building.

PR #48, opened 2020-01-27
[MSVC] added possible to compile at Microsoft Visual C compiler.

TheVice:
Hello.

In this pull request resolving compile errors at C compiler from Microsoft, described [here](https://github.com/BLAKE3-team/BLAKE3/issues/47).
If you are also interesting of linking, please also read [this issue](https://github.com/BLAKE3-team/BLAKE3/issues/16) - linking at MSVC environment have same issue like for the gcc.

Thank you.

### DIFF
--- a/c/main.c
+++ b/c/main.c
@@ -10,7 +10,6 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <unistd.h>
 
 #include "blake3.h"
 #include "blake3_impl.h"
--- a/c/blake3_avx2.c
+++ b/c/blake3_avx2.c
@@ -216,7 +216,7 @@ INLINE void transpose_msg_vecs(const uint8_t *const *inputs,
 
 INLINE void load_counters(uint64_t counter, bool increment_counter,
                           __m256i *out_lo, __m256i *out_hi) {
-  const __m256i mask = _mm256_set1_epi32(-(uint32_t)increment_counter);
+  const __m256i mask = _mm256_set1_epi32(-(int32_t)increment_counter);
   const __m256i add0 = _mm256_set_epi32(7, 6, 5, 4, 3, 2, 1, 0);
   const __m256i add1 = _mm256_and_si256(mask, add0);
   __m256i l = _mm256_add_epi32(_mm256_set1_epi32(counter), add1);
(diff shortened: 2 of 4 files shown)


c/blake3_neon.c
```diff
@@ -1,5 +1,5 @@
 #include "blake3_impl.h"
-
+#if !defined(_WIN32)
```
oconnor663: Why this change? Anything not explicitly targeting ARM NEON should be skipping this file entirely.

c/main.c
```diff
@@ -4,7 +4,9 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+#if !defined(_WIN32)
 #include <unistd.h>
```
oconnor663: Maybe we can remove this header entirely? It doesn't seem to affect the build when I try it.

At the same time, if we're adding Windows support here, we should probably figure out a way to add Windows CI testing in `.github/workflows/ci.yml`.

c/blake3_neon.c
TheVice: Guarding content of blake3_neon.c is recall/same approach used at request about ['using C code in C++'](https://github.com/BLAKE3-team/BLAKE3/pull/40). But in this case we have choose:
- even do not include this source file, through CMakeLists.txt, Visual Studio project, what ever other thing that able to build source files;
- or just guard file via macros - so no mater how we create project in this case.
It is no problem to not include this macros, because comparing to "external 'C'" construction such approach not so popular.

In case of "external 'C'" another approach will be that C++ side include C file inside proposed construction. 
```
extern "C" {
#include "blake3.h"
};
```

c/main.c
TheVice: Yes, for current content of main.c, including of this file can be removed. Probably it was left some experimentations that even not was committed.

Agree, adding of Windows CI testing environment would be right step to protect all changes proposed here and prevent regression changing that broke compile able at MSVC.

oconnor663:
One of the commits in this PR has a malformed commit message. A blank line between the summary line and the rest of the commit message is important, to avoid screwing up the formatting of `git log --oneline`. Could you please rebase and squash for commits, with a cleaned up message?

TheVice:
Sure.

Branch was re-based, commit are squashed into single one.

First one was created at TortoiseGit, second one at Git-Cola, where exists one-line control at the GUI.

oconnor663:
Thanks! Merging.

PR #49, opened 2020-01-27
[out] move creating/deleting of buffer named 'out' outside of 'do' loop.

TheVice:
Hello.

Pull request created for discussing at this [issue](https://github.com/BLAKE3-team/BLAKE3/issues/45) while talking at accompanying [request](https://github.com/BLAKE3-team/BLAKE3/pull/46).

Thank you.

### DIFF
--- a/c/main.c
+++ b/c/main.c
@@ -119,6 +119,13 @@ int main(int argc, char **argv) {
     assert(buf_len < buf_capacity);
   }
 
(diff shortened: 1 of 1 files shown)


oconnor663:
I think the code is simpler to understand as it is, and `main.c` is only for testing. Let's leave it.

PR #50, opened 2020-01-27
[WIP] [wide_buf] replace internal buffer at 'output_root_bytes' function with external.

TheVice:
Hello.

Pull request created for discussing at this [issue](https://github.com/BLAKE3-team/BLAKE3/issues/45).

Thank you.

### DIFF
--- a/c/blake3.c
+++ b/c/blake3.c
@@ -87,19 +87,19 @@ INLINE void output_chaining_value(const output_t *self, uint8_t cv[32]) {
 INLINE void output_root_bytes(const output_t *self, uint8_t *out,
                               size_t out_len) {
   uint64_t output_block_counter = 0;
-  uint8_t wide_buf[64];
+
   while (out_len > 0) {
     blake3_compress_xof(self->input_cv, self->block, self->block_len,
-                        output_block_counter, self->flags | ROOT, wide_buf);
-    size_t memcpy_len;
+                        output_block_counter, self->flags | ROOT, out);
+    size_t len;
     if (out_len > 64) {
-      memcpy_len = 64;
+      len = 64;
     } else {
-      memcpy_len = out_len;
+      len = out_len;
     }
-    memcpy(out, wide_buf, memcpy_len);
-    out += memcpy_len;
-    out_len -= memcpy_len;
+
+    out += len;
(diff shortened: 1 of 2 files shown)


c/blake3.c
```diff
@@ -87,19 +87,19 @@ INLINE void output_chaining_value(const output_t *self, uint8_t cv[32]) {
 INLINE void output_root_bytes(const output_t *self, uint8_t *out,
                               size_t out_len) {
   uint64_t output_block_counter = 0;
-  uint8_t wide_buf[64];
+
   while (out_len > 0) {
     blake3_compress_xof(self->input_cv, self->block, self->block_len,
-                        output_block_counter, self->flags | ROOT, wide_buf);
-    size_t memcpy_len;
+                        output_block_counter, self->flags | ROOT, out);
```
oconnor663: This is not legal. `blake3_compress_xof` always writes out 64 bytes, but the `out` buffer might be shorter than that.

What's your goal here?

TheVice: But why?
In it declaration it [requested 64 bytes](https://github.com/BLAKE3-team/BLAKE3/blob/163f52245d3f6cfcc2ec2c91e01412ffba5f6d45/c/blake3_dispatch.c#L228). Examining [portable variant](https://github.com/BLAKE3-team/BLAKE3/blob/a7579d30ad16c19306cedeeacd919e319aff1089/c/blake3_portable.c#L130) it operate with array of 16 of uint32_t that enough to store at 64 bytes according to sizeof of uint32_t to 4 bytes.

Back to wide_buf - it [too declared](https://github.com/BLAKE3-team/BLAKE3/blob/37ea737c16b5e0ad2909906e9e48856fd2e24f34/c/blake3.c#L90) as 64 bytes.

'For' at main make sure that all 'while' at blake3 loop step have enough space in buffer, so no need to call _memcpy_ at all in this case need, requested data already stored in our buffer.

From other side, caller response in this case that buffer have enough space, so previous approach will be more secure in this case, according to calls of this function not only from main file.
In this case, leaving original sources will be right choose.

oconnor663: I am having some trouble understanding your questions here.

PR #52, opened 2020-02-03
Inline wrapper methods

cesarb:


cesarb:
The checks failed due to a network error, and I cannot find a retry button (probably I don't have the correct permission), could you retry?

oconnor663:
Retrying. This sort of spurious failure happens more often than it should, maybe like 1 out of 10 builds. I wonder if I could configure the install step to retry a few times on its own?

oconnor663:
Thanks for catching these things. Is there an automated command you run to detect them? Maybe something I could add to CI?

PR #61, opened 2020-02-12
Rename assembly files (blake3-* -> blake3_*)

erijo:
This gives the assembly files the same prefix as the intrinsics files which
simplifies building when the build system should pick between the assembly and
the intrinsics files.

[Example usage](https://github.com/ccache/ccache/pull/519/commits/232809b46a5ae1a02e8fccfe1e4053edc5fe8d8d#diff-67e997bcfdac55191033d57a16d1408a)


oconnor663:
This seems like a good idea. I might land this, and then go a little further and make everything underscores. And maybe then turn "x86_64" into "x86-64", for good measure.

oconnor663:
Done and added fcc14c8c1b2891c0705ad5009078d0d618a1aa20. Thanks!

PR #64, opened 2020-02-15
create /tmp/bigfile with truncate instead of head

divinity76:
truncate will create a sparse file if supported by the system, which should rule out IO bottlenecks, and not actually consume 1 GB of disk space, and be much faster to create (especially if you're not running on a SSD) - unlike `head >`

- on my laptop with a Intel i7-9750H running in power-saving mode on battery with a 1TB Samsung 960 Evo SSD with a single-thread sequential-write performance of ~2076MB/s on CrystalMark 7:
```
$ time truncate -s 1GB bigfile

real    0m0.051s
user    0m0.000s
sys     0m0.030s

hans@LAPTOP-4AVGFSJI ~
$ rm bigfile

hans@LAPTOP-4AVGFSJI ~
$ time head -c 1000000000 /dev/zero > bigfile

real    0m1.269s
user    0m0.390s
sys     0m0.812s
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ hashing it, for example as follows:
 
 ```bash
 # Create a 1 GB file.
-head -c 1000000000 /dev/zero > /tmp/bigfile
+truncate -s 1GB /tmp/bigfile
 # Hash it with SHA-256.
 time openssl sha256 /tmp/bigfile
 # Hash it with BLAKE3.


oconnor663:
I like the idea, but does it work on macOS?

divinity76:
i don't know, i don't have a mac, but `dd` should be very portable right? you can do the same with dd, 
```
hans@LAPTOP-4AVGFSJI ~
$ time dd if=/dev/zero bs=1 count=0 seek=1GB of=bigfile
0+0 records in
0+0 records out
0 bytes copied, 0.0065183 s, 0.0 kB/s

real    0m0.054s
user    0m0.000s
sys     0m0.047s

hans@LAPTOP-4AVGFSJI ~
$ du --apparent-size --bytes bigfile
1000000000      bigfile
```

divinity76:
just googled it - seems `truncate` is NOT available in stock MacOS, but can be installed via `homebrew`, whatever that is.

oconnor663:
I think the current example with `head` is doing its job pretty well, so I'm going to close these.

PR #65, opened 2020-02-15
create /tmp/bigfile with dd instead of head

divinity76:
dd will create a sparse file if supported by the system, which should rule out IO bottlenecks, and not actually consume 1 GB of disk space, and be much faster to create (especially if you're not running on a SSD) - unlike `head >`

btw on GNU dd you can use `seek=1GB` instead of 1000000000, but not sure how portable that is on systems like MacOS or *BSD

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ hashing it, for example as follows:
 
 ```bash
 # Create a 1 GB file.
-head -c 1000000000 /dev/zero > /tmp/bigfile
+dd if=/dev/null bs=1 count=0 seek=1000000000 of=/tmp/bigfile
 # Hash it with SHA-256.
 time openssl sha256 /tmp/bigfile
 # Hash it with BLAKE3.


divinity76:
checked what would happen on a system that DOES NOT support sparse files:
```sh
# truncate -s 2G fat32.img
# mkfs.fat -F 32 fat32.img
mkfs.fat 4.1 (2017-01-24)
# mkdir fat32
# mount fat32.img fat32
# cd fat32
# dd if=/dev/null bs=1 count=0 seek=1000000000 of=bigfile1
0+0 records in
0+0 records out
0 bytes copied, 7.9143e-05 s, 0.0 kB/s
# dd if=/dev/null bs=1 count=0 seek=1000000000 of=bigfile2
0+0 records in
0+0 records out
0 bytes copied, 6.8841e-05 s, 0.0 kB/s
# dd if=/dev/null bs=1 count=0 seek=1000000000 of=bigfile3
dd: failed to truncate to 1000000000 bytes in output file 'bigfile3': No space left on device
#
```
looks good

oconnor663:
I'm a little wary of taking this change, because sparse files are kind of magical. I feel like a benchmark (even a silly informal one like this one) needs to serve two purposes, which are similar but not exactly the same:

1. Demonstrate that the thing you're measuring is fast.
2. Give you _confidence_ that the measurement is real and applies to your use case.

I think hashing a real file might be meaningfully better at (2). Users are more likely to know what the current command means, than they are to understand some of the less common `dd` flags, and that matters.

With that said, I'd also want to test this on macOS to see what happens. Has anyone gotten a chance to do that?

oconnor663:
I think the current example with `head` is doing its job pretty well, so I'm going to close these.

PR #81, opened 2020-04-23
allow for construction of hash in const fn context

inanna-malick:
This might be an antipattern that you don't want to encourage, but I found myself wanting to construct a const `Hash` in rust and added a shim to allow for it, eg:

```rust
pub const NULL_HASH: MyDomainHash = MyDomainHash(blake3::construct_magic_hash([0; 32]));
```

Happy to add more documentation if this is something you'd be open to merging.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -198,6 +198,12 @@ fn counter_high(counter: u64) -> u32 {
 #[derive(Clone, Copy, Hash)]
 pub struct Hash([u8; OUT_LEN]);
 
+
+/// Provided to allow for construction of magic hash values (eg null commit in git as [0; OUT_LEN])
+pub const fn construct_magic_hash(bytes: [u8; OUT_LEN]) -> Hash {
+    Hash(bytes)
+}
+
 impl Hash {
     /// The bytes of the `Hash`. Note that byte arrays don't provide
     /// constant-time equality checking, so if  you need to compare hashes,


cesarb:
There's already `Hash::from([0; 32])`, so creating a `Hash` directly has always been allowed (and I see no reason why it would be discouraged). It seems trait fns cannot be const, but since this is already allowed, I would recommend naming it simply `Hash::new(...)`.

oconnor663:
I like @cesarb's suggestion of `Hash::new`. The main reason we haven't had it before is just that it's redundant with `From`, but if `From` can't be `const` then that's a good reason to have `new` also. There's some risk that a caller might confuse `blake3::hash()` (actually hash some bytes) with `blake3::Hash::new()` (just cast some bytes into a `Hash` output), but I think the `[u8; 32]` argument type makes that risk pretty small.

oconnor663:
It looks like we might be at a place where the entire hash function could be implemented with `const fn`? Here's some similar work: https://www.reddit.com/r/rust/comments/hi11op/announcing_constsha1_a_sha1_implementation_for/

AZon8:
I translated the  reference_impl into const, but it requires rewriting the for loops and doing manual indexing. It's not very readable. Hopefully within a year it can be const-ified as is.

In case anybody is interested:

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018

oconnor663:
@AZon8 that playground link doesn't include your code. You might need to click the "Share" button.

AZon8:
Whoops. 

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=bf14a7fc359a5a072b3f8df690b593dc

Issue #84, opened 2020-05-04
Support re-creating Hasher state


That3Percent:
I would like to pause and resume the hashing of a long-lived sequence of events that are collected over time by saving a `Hasher` instance to a database.

There are a couple of ways this could be accomplished API-wise:

1. (preferred) `serde` integration
2. Constructor that takes arguments made available by getter functions.

If this is something that the authors would be amenable to adding, I'd be willing to help out with a PR but may need some guidance about what to potentially be concerned about since I'm not familiar with the source or how much of the implementation details to expose.

oconnor663:
Interesting. I'm a little hesitant to add a serialization mechanism that's as "automatic" as Serde. Like most modern hash functions, BLAKE3 is designed to be resistant to length extension, to prevent a large class of attacks. If you can deserialize a `Hasher`, you can also length-extend whatever hash it produced. For many applications that's no big deal, because their input isn't secret, but for other applications it's as sensitive as a copy of their secret key (in fact it includes a copy of their secret key). Knowing which category you're in isn't always easy. `Hasher` implements `Debug` manually for similar reasons, because `#[derive(Debug)]` could unintentionally leak secrets.

What have other hash implementations done about this sort of thing? It looks like e.g. [`ring::digest::Context`](https://briansmith.org/rustdoc/ring/digest/struct.Context.html) doesn't support exporting.

Some other scattered thoughts:

In practice, serializing a `Hasher` by transmuting its bytes with unsafe code should work across processes spawned from the same binary. It's a `no_std`-compatible type, without any heap allocation. That said, Rust makes no guarantees about things like field ordering (since it's not a `repr(C)` type), so this wouldn't be safe if you wanted to reconstitute the bytes in a future version of your app, or on a different platform. I think there might also be UB issues associated with reading padding bytes? Anyway, I'm not recommending doing this, but if you're life depends on serializing a `Hasher` right this second, it might work :)

[...]

PR #85, opened 2020-05-06
add export_hasher and import_hasher to the guts module

oconnor663:
This was requested in https://github.com/BLAKE3-team/BLAKE3/issues/84.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,8 +41,8 @@ jobs:
     - name: print instruction set support
       run: cargo run --quiet
       working-directory: ./tools/instruction_set_support
-    # Default tests plus Rayon.
-    - run: cargo test --features=rayon
+    # Default tests plus Rayon and Serde.
+    - run: cargo test --features=rayon,serde_derive
     # no_std tests.
     - run: cargo test --no-default-features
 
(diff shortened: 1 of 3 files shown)


oconnor663:
This is an experimental PR. @That3Percent, if you have time to try this out for your use case, I'd love to get feedback.

That3Percent:
Nice!

It will be at least a few days before I can have an end to end test. I reviewed the changes though and the API and functionality needed are confirmed by the unit tests, so I have every expectation that this will work great.

Cheers,
-Zac

That3Percent:
We ended up going with a different solution that still uses Blake3 under the hood (especially for the XOF, which is great!). Instead of trying to transform all of the data into a contiguous byte stream and save and load hasher state (which was a pain because of the linear nature of hashes over byte streams compared to the nature of our problem) instead go with a solution that allowed us to treat all of the data as independent parts of a very large struct that can be hashed in parallel and combined piecewise.

You can read a little bit about what we're doing now here: [stable structured hash](https://www.reddit.com/r/crypto/comments/gjwmn4/please_break_my_stable_structured_hash_function/)

I feel really bad about you having implemented a thing we didn't use. Very sorry about that. It wasn't clear that this massively parallel solution was even possible until after.

oconnor663:
No worries! This is an interesting experiment, and we can look back at this PR if we ever need it in the future.

> stable structured hash

Sounds complicated! :sweat_smile:

iajoiner:
@oconnor663 Actually I have a use case for it.

PR #90, opened 2020-06-12
Bump digest crypto-mac

### LINKED ISSUE #89: Bump digest to v0.9; crypto-mac to v0.8
Hello! Thanks for BLAKE3!

Heads up, we've released new versions of `digest` and `crypto-mac`. They now use `generic-array` v0.14 with an MSRV of 1.41+. We've also moved to Initialize-Update-Finalize (IUF) terminology within the API.

Release notes:
- `digest` v0.9: https://github.com/RustCrypto/traits/pull/186
- `crypto-mac` v0.8: https://github.com/RustCrypto/traits/pull/169

Stupremee:
This PR bumps digest to 0.9.0 and crypto-mac to 0.8.0

Resolves #89

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,8 +75,8 @@ arrayvec = { version = "0.5.1", default-features = false, features = ["array-siz
 constant_time_eq = "0.1.5"
 rayon = { version = "1.2.1", optional = true }
 cfg-if = "0.1.10"
-digest = "0.8.1"
-crypto-mac = "0.7.0"
+digest = "0.9.0"
+crypto-mac = "0.8.0"
 
 [dev-dependencies]
 hex = "0.4.2"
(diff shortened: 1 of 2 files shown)


oconnor663:
Thank you! I've followed up with 4c41a893a00a3ebe7b24529531ccf96d8593a57c. As far as releasing this, do you think it should be a backwards-incompatible version bump? What do most other crates do?

Stupremee:
Imo there's no need to be backwards-compatible, because if the user really needs <0.9.0, he can just use an older version of `blake3`

oconnor663:
My worry is that an automatic upgrade to the next version of `blake3` could break some callers, if they're relying on the traits we re-export? Maybe this is an argument for moving the trait implementations out to a wrapper crate (defining a trivial wrapper type to implement traits on), where it would make more sense to do a backwards-incompatible version bump every time the trait does? Not sure.

Stupremee:
That would make sense yes. Then there could be different versions of the extra crate that all implement the traits for different versions of the `digest` crate.

Issue #21, opened 2020-01-13
enable c_neon by default on aarch64


oconnor663:
NEON is always supported on Aarch64, so there's no need to make it conditional on the `c_neon` flag. I don't think `build.rs` can enable features, so we might have to do this with `target_arch = "aarch64" OR feature = "c_neon"` everywhere, which is doable but a bit unfortunate. (Not so different from how we have to say `x86 OR x86_64` all over the place.)

PR #91, opened 2020-06-20
Neon detection

xnox:


xnox:
Hm... actually cpu_feature has to be defined on per-architecture basis, because one cannot cycle through AVX on ARM, and vice versa.

sneves:
The way we iterate though features in `main.c` will only cycle through available features:
```
const int mask = get_cpu_features();
int feature = 0;
do {
  ...
  feature = (feature - mask) & mask;
} while(feature != 0);
```
So, in ARM it will be only 0 and potentially NEON.

xnox:
> The way we iterate though features in `main.c` will only cycle through available features:
>
Ack.
 
> So, in ARM it will be only 0 and potentially NEON.

Yeap. For now at least.

But we could grow SVE, SV2 too. And maybe Atomics usage might help too (as seen on AWS Graviton2 instances).

oconnor663:
Nice! This would close out https://github.com/BLAKE3-team/BLAKE3/issues/21. I think statically enabling NEON for AArch64 is great. Does AArch32 also always support NEON/ASIMD, and is there any way for us to detect AArch32 in stable Rust?

I'm not very familiar with ARM stuff in general, but my impression [from StackOverflow](https://stackoverflow.com/a/26920837/823869) is that there's no OS-independent way to detect NEON support on ARMv7. If I'm reading correctly, it looks like here we're only doing dynamic detection on Linux. Is that also going to work on Android? (The SO answer linked there suggests that Android needs a different method.) On the other hand, maybe Android is always going to be AArch64 anyway these days?

xnox:
> Nice! This would close out https://github.com/BLAKE3-team/BLAKE3/issues/21. I think statically enabling NEON for AArch64 is great. Does AArch32 also always support NEON/ASIMD, and is there any way for us to detect AArch32 in stable Rust?
> 
AArch32 is marketing name for all of old 32bit arm ports. And one still has to detect NEON/ASIMD there.

On Linux, one can use getauxval method from Rust too.

> I'm not very familiar with ARM stuff in general, but my impression [from StackOverflow](https://stackoverflow.com/a/26920837/823869) is that there's no OS-independent way to detect NEON support on ARMv7. If I'm reading correctly, it looks like here we're only doing dynamic detection on Linux. Is that also going to work on Android? (The SO answer linked there suggests that Android needs a different method.) On the other hand, maybe Android is always going to be AArch64 anyway these days?

I believe Android ndk v18 provides getauxval API too.

So 64bit ARM should be covered everywhere. 32bit ARM is fixed on classic Linux, recentish Androids.

Windows/MacOS are mostly 64bit ARM these days.

And I still need to fixup more of rust code to do the same.

oconnor663:
I have a collection of Raspberry Pis sitting around that I can test this on. Let me know when this is ready for a lot of manual testing. Also apologies in advance for being slower to respond these days -- I'm in a busy period with a new job.

src/platform.rs
```diff
@@ -351,6 +354,39 @@ pub fn sse41_detected() -> bool {
...
+    // getauxval, yet it's not available. So use libc ffi bindings to
+    // get to getauxval ourselves.
+    // #define HWCAP_ARM_VFP 64
+    // #define HWCAP_ARM_NEON 4096
+    //
+        unsafe {
+            return (libc::getauxval(libc::AT_HWCAP) & (64 | 4096)) != 0
+        }
+    }
+}
```
xnox, line 388: This implements neon_detect the same way as done in the C implementation.

Note this is my first time writing rust code. So please review this carefully.

Namely it's simply true on aarch64, and on armhf an unsafe call to libc getauxval is made. It is implemented by Android's bionic library & glibc.

I have slight concerns that it might not be portable to FreeBSD/MacOS. However, I don't think there are any 32-bit ARM MacOS things left. But FreeBSD is a concern.

For example, I have no idea how above behaves if `AT_HWCAP` is not defined, or `getauxval` is not defined in the target libc. I hope that compilation fails, when building against such libc.

c/blake3_c_rust_bindings/build.rs
```diff
@@ -122,10 +127,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         avx512_build.compile("blake3_avx512");
     }
 
-    // We only build NEON code here if 1) it's requested and 2) the root crate
-    // is not already building it. The only time this will really happen is if
-    // you build this crate by hand with the "neon" feature for some reason.
-    if defined("CARGO_FEATURE_NEON") {
+    // We only build NEON code here if 1) on arm64 or it's requested and 2) the root crate
+    // is not already building it.
+    if is_aarch64() || defined("CARGO_FEATURE_NEON") {
```
xnox, line 132: Needs fixing, feature no longer exists.

xnox:
Completing dropping NEON brought up a few compile issues.

Changing VFP detection to VFPv4, as that's what is needed for NEON.

And I feel that portable implementation should still be testable on AArch64. So I am declaring NEON feature, despite required, such that test harness can test portable and neon implementations. Same as SSE on x86_64.

sneves:
On the C side, this seems to work OK on Linux. I made the following change to cycle though reference and NEON code when testing:
```diff
diff --git a/c/blake3_impl.h b/c/blake3_impl.h
index 98272d4..223fffd 100644
--- a/c/blake3_impl.h
+++ b/c/blake3_impl.h
@@ -73,8 +73,8 @@ enum cpu_feature {
   AVX512F = 1 << 5,
   AVX512VL = 1 << 6,
 #endif
-#if defined(IS_ARMHF)
-  NEON = 1 << 0,
+#if defined(IS_ARM)
+  NEON = 1 << 7,
 #endif
   /* ... */
   UNDEFINED = 1 << 30
```
I can't comment on the Rust side.

xnox:
> On the C side, this seems to work OK on Linux/AArch64. I made the following change to cycle though reference and NEON code when testing:
> 
> ```diff
> diff --git a/c/blake3_impl.h b/c/blake3_impl.h
> index 98272d4..223fffd 100644
> --- a/c/blake3_impl.h
> +++ b/c/blake3_impl.h
> @@ -73,8 +73,8 @@ enum cpu_feature {
>    AVX512F = 1 << 5,
>    AVX512VL = 1 << 6,
>  #endif
> -#if defined(IS_ARMHF)
> -  NEON = 1 << 0,
> +#if defined(IS_ARM)
> +  NEON = 1 << 7,
>  #endif

The define change is correct. The bump from 0 to 7, is not. As none of the other optimizations are available on ARM, as NEON is the first level of optimisation on.

On ARM, I expect `1<<1` be defined as SVE, `1<<2` as SVE2, etc.

sneves:
I would prefer feature bits not to overlap, even across separate architectures, but that's your call.

`c/blake3_c_rust_bindings/src/lib.rs` needs a definition of `neon_detected`.

xnox:
I'm pushing my work-in-progress code, it fixes a few things, but I think it's still incomplete. Did not finish testing that v6, v7, v8 arm builds and tests correctly across rust & c.

xnox:
> I would prefer feature bits not to overlap, even across separate architectures, but that's your call.

That's ok, but then can there be delimiter of min features? such that one iterates implementations from min to max, for a given arch? such that e.g. 7 is arm portable, and 8 is arm neon, and the C test program does not iterate through all the avx stuff (and like exercising portable implementation many times).

sneves:
Maybe I'm misunderstanding what you're saying, but the test program does not iterate through AVX etc. Say, for example, you have an ARM chip with NEON and SVE2 extensions: https://gcc.godbolt.org/z/J8i6MV

The program will iterate only through the combination of features NONE, NEON|NONE, SVE2|NONE, SVE|NEON|NONE, and ignore all other values (here NONE = 0 = portable code).

xnox:
> Maybe I'm misunderstanding what you're saying, but the test program does not iterate through AVX etc. Say, for example, you have an ARM chip with NEON and SVE2 extensions: https://gcc.godbolt.org/z/J8i6MV
> 
> The program will iterate only through the combination of features NONE, NEON|NONE, SVE2|NONE, SVE|NEON|NONE, and ignore all other values (here NONE = 0 = portable code).

i misunderstood how it works. And that website is really nice to show C code!

Cargo.toml
```diff
@@ -12,12 +12,6 @@ edition = "2018"
 [features]
 default = ["std"]
 
-# The NEON implementation does not participate in dynamic feature detection,
-# which is currently x86-only. If "neon" is on, NEON support is assumed. Note
-# that AArch64 always supports NEON, but support on ARMv7 varies. The NEON
-# implementation uses C intrinsics and requires a C compiler.
-neon = []
```
nvzqz, line 19: If this change lands, something to keep in mind is that removing Cargo features is considered a breaking change and would require a `0.x` to `0.y` version bump. It might be better to release a backward-compatible version with NEON detection, and then later remove the flag when other breaking changes are happening.

b8591340:
What's missing here for it to be merged? I have been using it on iOS without issues.

PR #92, opened 2020-06-20
c: Implement ifunc based dispatcher

xnox:
ifunc based dispatcher is a linux/glibc/ELF feature which allows to
run detection code at early elf startup, and bind a particular
implementation for a function. After that, the chosen implementation
will be executed when refer to by the the common name.

Added an option to use the ifunc dispatcher for any Makefile target,
simply specify BLAKE3_DISPATCH=ifunc, like so:

make test_asm CFLAGS='-DBLAKE3_TESTING=1' BLAKE3_DISPATCH=ifunc
make test CFLAGS='-DBLAKE3_TESTING=1' BLAKE3_DISPATCH=ifunc

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -5,6 +5,12 @@ TARGETS=
 ASM_TARGETS=
 EXTRAFLAGS=
 
+ifeq (ifunc,$(BLAKE3_DISPATCH))
+DISPATCH=blake3_dispatch_ifunc.c
+else
+DISPATCH=blake3_dispatch.c
+endif
+
 ifdef BLAKE3_NO_SSE41
 EXTRAFLAGS += -DBLAKE3_NO_SSE41
 else
@@ -31,7 +37,7 @@ EXTRAFLAGS += -DBLAKE3_USE_NEON
 TARGETS += blake3_neon.o
 endif
 
-all: blake3.c blake3_dispatch.c blake3_portable.c main.c $(TARGETS)
+all: blake3.c $(DISPATCH) blake3_portable.c main.c $(TARGETS)
 	$(CC) $(CFLAGS) $(EXTRAFLAGS) $^ -o $(NAME)
 
 blake3_sse41.o: blake3_sse41.c
@@ -50,7 +56,7 @@ test: CFLAGS += -DBLAKE3_TESTING -fsanitize=address,undefined
 test: all
 	./test.py
 
-asm: blake3.c blake3_dispatch.c blake3_portable.c main.c $(ASM_TARGETS)
+asm: blake3.c $(DISPATCH) blake3_portable.c main.c $(ASM_TARGETS)
 	$(CC) $(CFLAGS) $(EXTRAFLAGS) $^ -o $(NAME)
 
 test_asm: CFLAGS += -DBLAKE3_TESTING -fsanitize=address,undefined
--- a/c/README.md
+++ b/c/README.md
@@ -214,6 +214,15 @@ in call to always_inline ‘vaddq_u32’: target specific option mismatch
 ...then you may need to add something like `-mfpu=neon-vfpv4
 -mfloat-abi=hard`.
 
+## Linux ifunc dispatcher
+
+Instead of the regular/portable dispatcher, on Linux one can use ifunc
+dispatcher. With ifunc based dispatcher the best versions of the
+functions are resolved at startup time and used throughout the
+execution. Simply use `blake3_dispatch_ifunc.c` instead of
+`blake_dispatch.c`. Or specify `BLAKE3_DISPATCH=ifunc` to the `make`
+command.
+
 ## Other Platforms
 
 The portable implementation should work on most other architectures. For
--- a/c/blake3_dispatch_ifunc.c
+++ b/c/blake3_dispatch_ifunc.c
@@ -1,0 +1,119 @@
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include "blake3_impl.h"
+
+#if defined(IS_ARMHF) && defined(__linux__)
+#include <sys/auxv.h>
+#include <asm/hwcap.h>
+#endif
+
+#if defined(IS_ARMHF)
+#define IFUNC_RESOLVER_ARGS uint64_t hwcap
+#else
+#define IFUNC_RESOLVER_ARGS void
+#endif
+
+void blake3_compress_in_place(uint32_t cv[8],
+                              const uint8_t block[BLAKE3_BLOCK_LEN],
+                              uint8_t block_len, uint64_t counter,
+                              uint8_t flags) __attribute__((ifunc ("resolve_compress_in_place")));
+
+typedef void (*f_blake3_compress_in_place)(uint32_t *, const uint8_t *, uint8_t, uint64_t, uint8_t);
+static f_blake3_compress_in_place resolve_compress_in_place(IFUNC_RESOLVER_ARGS) {
+#if defined(IS_X86)
+  __builtin_cpu_init();
+  if (__builtin_cpu_supports("avx512vl"))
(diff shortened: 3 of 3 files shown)


oconnor663:
@sneves I'll defer to you on this one. @xnox have you been able to measure a performance improvement from using ifuncs? I've always assumed that our dispatching overhead was in the noise, but I don't have much data on it. (Compiling with RUSTFLAGS="-C target-cpu=native" on my AVX2 laptop does show a few cycles of improvement per compression, but I don't know how many other effects it has besides optimizing out the dispatch.)

xnox:
> @sneves I'll defer to you on this one. @xnox have you been able to measure a performance improvement from using ifuncs? I've always assumed that our dispatching overhead was in the noise, but I don't have much data on it. (Compiling with RUSTFLAGS="-C target-cpu=native" on my AVX2 laptop does show a few cycles of improvement per compression, but I don't know how many other effects it has besides optimizing out the dispatch.)

In the ifunc dispatcher I used builtins for CPU features detection. I wonder if I can bring that over to the regular dispatcher, instead of using hand coded cpuid calls. Or at least use `cpuid.h` on Linux there.

I will try to instrument measurements for this. In theory, this should only help very little for long running processes as dispatcher code is only executed once per application lifetime. Instead of on every invocation.

oconnor663:
> as dispatcher code is only executed once per application lifetime. Instead of on every invocation.

Might be worth clarifying, our dispatcher code today only executes once per process. Subsequent calls use the cached result: https://github.com/BLAKE3-team/BLAKE3/blob/master/c/blake3_dispatch.c#L85-L87

sneves:
> In the ifunc dispatcher I used builtins for CPU features detection. I wonder if I can bring that over to the regular dispatcher, instead of using hand coded cpuid calls. Or at least use cpuid.h on Linux there.

We don't target GCC/Linux exclusively. Using builtins just means we'd have to maintain the builtin version and the other version.

I considered at the time using function pointers instead of regular branches for dispatching, as I had done before on BLAKE2. One, perhaps irrelevant, scenario where it matters is if you enable Spectre retpoline mitigations (e.g. `-mretpoline, -mindirect-branch=thunk`); then your function pointer overhead spikes by ~70 cycles, whereas the current dispatch method remains more or less constant. But this doesn't matter when using ifuncs, since they're implemented as relocations.

Anyway, back to overhead. I'm not sure there's a measurable benefit. I wrote a specific benchmark to measure the overhead of a single `blake3_compress_in_place` call, which would be the most affected by this. The results, on a Skylake chip and the median of 2^24 measurements, are thus:

Current dispatching:

    blake3_compress_in_place: 240 cycles/compression
    blake3_compress_in_place_sse41: 234 cycles/compression

ifunc dispatching:

    blake3_compress_in_place: 240 cycles/compression
    blake3_compress_in_place_sse41: 234 cycles/compression

So performance-wise, it doesn't seem to make much of a difference. There's around 6 cycles of overhead relative to calling the function directly.

However, if we enable LTO, the detection code of `blake3_compress_in_place` can get inlined into the loop and the function call overhead is reduced:

    blake3_compress_in_place: 238 cycles/compression

This is not the case for ifunc, which always remains at 240.

In any case, since this is mostly a very contained feature, I don't have much of a problem with including the ifunc dispatcher and letting users choose.

xnox:
> > In the ifunc dispatcher I used builtins for CPU features detection. I wonder if I can bring that over to the regular dispatcher, instead of using hand coded cpuid calls. Or at least use cpuid.h on Linux there.
> 
> We don't target GCC/Linux exclusively. Using builtins just means we'd have to maintain the builtin version and the other version.
> 
There will always be platform specific code. `__builtin_cpu_supports` is supported by GCC and CLANG, which gives coverage across MacOS/FreeBSD/Android/Linux, and Windows when those compilers are used. And a different code path is already used for Windows. What I am proposing is portable and is simplier to use and understand. And can be extended to more optimisations later.


> I considered at the time using function pointers instead of regular branches for dispatching, as I had done before on BLAKE2. One, perhaps irrelevant, scenario where it matters is if you enable Spectre retpoline mitigations (e.g. `-mretpoline, -mindirect-branch=thunk`); then your function pointer overhead spikes by ~70 cycles, whereas the current dispatch method remains more or less constant. But this doesn't matter when using ifuncs, since they're implemented as relocations.
> 
> Anyway, back to overhead. I'm not sure there's a measurable benefit. I wrote a specific benchmark to measure the overhead of a single `blake3_compress_in_place` call, which would be the most affected by this. The results, on a Skylake chip and the median of 2^24 measurements, are thus:
> 
> Current dispatching:
> 
> ```
> blake3_compress_in_place: 240 cycles/compression
> blake3_compress_in_place_sse41: 234 cycles/compression
> ```
> 
> ifunc dispatching:
> 
> ```
> blake3_compress_in_place: 240 cycles/compression
> blake3_compress_in_place_sse41: 234 cycles/compression
> ```
> 
> So performance-wise, it doesn't seem to make much of a difference. There's around 6 cycles of overhead relative to calling the function directly.
> 
> However, if we enable LTO, the detection code of `blake3_compress_in_place` can get inlined into the loop and the function call overhead is reduced:
> 
> ```
> blake3_compress_in_place: 238 cycles/compression
> ```
> 
> This is not the case for ifunc, which always remains at 240.
> 

Can you share the benchmark you did to detect this? This is very interesting analysis which might be helpful in other code too. Especially the effect of lto on dispatcher.

> In any case, since this is mostly a very contained feature, I don't have much of a problem with including the ifunc dispatcher and letting users choose.

Thanks. But I'd want to reduce the number of compile options eventually, to publish a performant shared C library, on x86, arm, power, and z.

sneves:
```
#include <cstddef>
#include <cstdint>
#include <cstdlib>
#include <cstdio>

#include <vector>
#include <algorithm>

#include <immintrin.h>

extern "C" {
	void blake3_compress_in_place(uint32_t cv[8], const uint8_t block[64], uint8_t block_len, uint64_t counter, uint8_t flags);
	void blake3_compress_in_place_sse41(uint32_t cv[8], const uint8_t block[64], uint8_t block_len, uint64_t counter, uint8_t flags);
}

namespace {

	static inline uint64_t rdtsc() {
	  _mm_lfence();
	#if defined(__clang__)
	  return __builtin_readcyclecounter();
	#else
	  uint32_t hi, lo = 0;
	  __asm__ __volatile__("rdtscp" : "=a"(lo), "=d"(hi)::"rcx");
	  return lo | (((uint64_t)hi) << 32);
	#endif
	}

	template<typename Function>
	static inline uint64_t bench(Function F, size_t const iters) {
		const uint8_t block[64] = {};
		uint32_t cv[8] = {};
		std::vector<uint64_t> cycles(iters + 1);
		for(auto& c : cycles) {
			c = rdtsc();
			F(cv, block, 64, 123, 0x99);
			 __asm__ __volatile__("" : : "m"(cv), "m"(block) : "memory");
		}
		for (size_t i = 0; i < iters; ++i) {
	    cycles[i] = cycles[i + 1] - cycles[i];
	  }
	  cycles.pop_back();
	  std::sort(begin(cycles), end(cycles));
	  return cycles[iters / 2];
	}

}

int main() {
	const size_t kIters = 1UL << 24;
	printf("blake3_compress_in_place: %lu\n", bench([](auto...args) { blake3_compress_in_place(args...); }, kIters));
	printf("blake3_compress_in_place_sse41: %lu\n", bench([](auto...args) { blake3_compress_in_place_sse41(args...); }, kIters));
	return 0;
}
```

c/Makefile
```diff
@@ -5,6 +5,12 @@ TARGETS=
 ASM_TARGETS=
 EXTRAFLAGS=
 
+ifeq (ifunc,$(BLAKE3_DISPATCH))
```
xnox, line 8: This build variant must be added to the Github actions test cases.

Issue #95, opened 2020-06-25
Intel CET missing for asm implementation


xnox:
Intel Control-flow Enforcement Technology (CET):

https://software.intel.com/en-us/articles/intel-sdm

Using CET enabled toolchain (ie. Ubuntu Focal or Groovy), CET is enabled for intrinsic implementation, but not the assembly.

```
$ make test
$ readelf -a blake3 | grep IBT
      Properties: x86 feature: IBT, SHSTK
$ make clean
$ make test_asm
$ readelf -a blake3 | grep IBT
$
# no output, thus no CET support
```

I suspect that assembly files must include `cet.h` header and have endbranch declaration similar to e.g. other asm implementations in other software.

endbranch declarations needed & the gnu.property:
https://github.com/openssl/openssl/pull/12272/files

The gnu.property can also be added by including a cet header:
https://github.com/gpg/libgcrypt/commit/4c88c2bd2a418435506325cd53246acaaa52750c

PR #96, opened 2020-06-26
Assembly: enable CET

sneves:
Addresses #95.

With our current dispatcher, both on C and Rust, having `endbr64` at the top of the functions is rather pointless since these will never be at the end of an indirect call. But since this is a glorified `nop`, might as well.

### DIFF
--- a/c/blake3_avx2_x86-64_unix.S
+++ b/c/blake3_avx2_x86-64_unix.S
@@ -1,3 +1,9 @@
+#if defined(__ELF__) && defined(__CET__) && __has_include(<cet.h>)
+#include <cet.h>
+#else
+#define _CET_ENDBR
+#endif
+
 .intel_syntax noprefix
 .global _blake3_hash_many_avx2
 .global blake3_hash_many_avx2
@@ -9,6 +15,7 @@
         .p2align  6
 _blake3_hash_many_avx2:
 blake3_hash_many_avx2:
+        _CET_ENDBR
         push    r15
         push    r14
         push    r13
--- a/c/blake3_avx512_x86-64_unix.S
+++ b/c/blake3_avx512_x86-64_unix.S
@@ -1,5 +1,10 @@
-.intel_syntax noprefix
+#if defined(__ELF__) && defined(__CET__) && __has_include(<cet.h>)
+#include <cet.h>
+#else
+#define _CET_ENDBR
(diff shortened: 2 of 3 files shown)


oconnor663:
I have no idea what this means. Can I leave it to you to hit the merge button whenever you think it's ready? :)

xnox:
Looks good to me! The test C binary is marked as CET enabled.

Please merge.

xnox:
> I have no idea what this means. Can I leave it to you to hit the merge button whenever you think it's ready? :)

I find this presentation the best explanation about the CET technology https://www.linuxplumbersconf.org/event/2/contributions/147/attachments/72/83/CET-LPC-2018.pdf

xnox:
Maybe @hjl-tools could review this too. This adds CET support to BLAKE3 asm.

hjl-tools:
Since the assembly codes are in .S,  you should do

```
#if defined(__ELF__) && defined(__CET__)
# include <cet.h>
#else
# define _CET_ENDBR
#endif
```
and use _CET_ENDBR everywhere ENDBR is needed.  Since -fcf-protection needs binutils with CET support,
it should work.

sneves:
The reason I did not just include the `cet.h` header is that it appears to be GCC-specific; building with Clang would fail, even though it supports `-fcf-protection=full`.

hjl-tools:
> The reason I did not just include the `cet.h` header is that it appears to be GCC-specific; building with Clang would fail, even though it supports `-fcf-protection=full`.

LLVM doesn't working CET support before LLVM 11 which also includes <cet.h>:

https://bugs.llvm.org/show_bug.cgi?id=45484

Most of CET fixes have been backported to LLVM 10.x, except for <cet.h>.   You can do

```
#if defined(__ELF__) && defined(__CET__) && __has_include(<cet.h>)
# include <cet.h>
#else
...
#endif
```

sneves:
> Most of CET fixes have been backported to LLVM 10.x, except for <cet.h>. You can do
> 
> ```
> #if defined(__ELF__) && defined(__CET__) && __has_include(<cet.h>)
> # include <cet.h>
> #else
> ...
> #endif
> ```

Goes in as suggested.

xnox:
Thank you @hjl-tools for your input! Much appreciated!

PR #100, opened 2020-07-23
feat: Implement `core::hash::Hasher` for `Hasher`

Hywan:
This PR is an attempt to implement `core::hash::Hasher` for `Hasher`.

To achieve that, I first needed to transform `Hash` into a `u64` (https://github.com/BLAKE3-team/BLAKE3/commit/bde6e6f8918b1d909ea4e1a006da5f893a279dbf):

> This is an attempt to encode an `Hash` into `u64`. This transformation
will increase the number of collisions as `u64` can represent fewer
values than `[u8; OUT_LEN]` (at the time of writing, `OUT_LEN` is set
to 32).
>
> Bytes from `Hash` are extracted by chunks of 8, and then added to
`u64` with a modular addition.

Then, `core::hash::Hasher` can be implemented (https://github.com/BLAKE3-team/BLAKE3/commit/41baa2f7759c95d1b8469639cfc02c8946a1dfb3):

> This is an attempt to implement the standard `Hasher` API for
`blake3`, so that implementors can use `blake3` when a generic hasher
is expected.

Things that must be discussed:

1. Do you want `Hasher` to implement `core::hash::Hasher`?
2. Do you find acceptable the strategy to convert `Hash` into `u64`?
3. If `OUT_LEN % 8 != 0`, bytes will be missed by `From<Hash>` for `u64`. Can it be different? Do we have to assert that?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1266,6 +1266,17 @@ impl std::io::Write for Hasher {
     }
 }
 
+#[cfg(feature = "std")]
+impl std::hash::Hasher for Hasher {
+    fn write(&mut self, bytes: &[u8]) {
+        self.update(bytes);
+    }
+
+    fn finish(&self) -> u64 {
+        self.finalize().into()
+    }
+}
+
 /// An incremental reader for extended output, returned by
 /// [`Hasher::finalize_xof`](struct.Hasher.html#method.finalize_xof).
 #[derive(Clone)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -239,6 +239,37 @@ impl From<Hash> for [u8; OUT_LEN] {
     }
 }
 
+/// Transforming `Hash` into `u64` will increase the number of
+/// collisions as `u64` can represent fewer values than `[u8;
+/// OUT_LEN]`.
+///
+/// Bytes from `Hash` are extracted by chunks of 8, and then added to
+/// `0u64` with a modular addition.
+#[cfg(feature = "std")]
+impl From<Hash> for u64 {
+    #[inline]
+    fn from(hash: Hash) -> Self {
+        use std::convert::TryInto;
+        use std::u64;
+
+        let bytes: [u8; OUT_LEN] = hash.into();
+
+        (0..(OUT_LEN / 8))
+            .into_iter()
+            .map(|i| (i * 8, (i + 1) * 8))
+            .fold(0u64, |accumulator, (x, y)| {
+                assert_eq!(y - x, 8);
+
+                accumulator.wrapping_add(u64::from_be_bytes(
+                    (&bytes[x..y])
+                        .try_into()
+                        // SAFETY: OK, because `y - x = 8` and `wrapping_add` expects a `[u8; 8]`.
+                        .expect("failed to convert `&[u8]` into `[u8; 8]`"),
+                ))
+            })
+    }
+}
+
 /// This implementation is constant-time.
 impl PartialEq for Hash {
     #[inline]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -245,12 +245,10 @@ impl From<Hash> for [u8; OUT_LEN] {
 ///
 /// Bytes from `Hash` are extracted by chunks of 8, and then added to
 /// `0u64` with a modular addition.
-#[cfg(feature = "std")]
 impl From<Hash> for u64 {
     #[inline]
     fn from(hash: Hash) -> Self {
-        use std::convert::TryInto;
-        use std::u64;
+        use core::convert::TryInto;
 
         let bytes: [u8; OUT_LEN] = hash.into();
 
@@ -1266,8 +1264,7 @@ impl std::io::Write for Hasher {
     }
 }
 
-#[cfg(feature = "std")]
-impl std::hash::Hasher for Hasher {
+impl core::hash::Hasher for Hasher {
     fn write(&mut self, bytes: &[u8]) {
         self.update(bytes);
     }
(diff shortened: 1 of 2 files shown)


oconnor663:
There was a similar proposal here, but we decided against it: https://github.com/BLAKE3-team/BLAKE3/pull/87

src/lib.rs
```diff
@@ -239,6 +239,35 @@ impl From<Hash> for [u8; OUT_LEN] {
...
+        (0..(OUT_LEN / 8))
+            .into_iter()
+            .map(|i| (i * 8, (i + 1) * 8))
+            .fold(0u64, |accumulator, (x, y)| {
+                assert_eq!(y - x, 8);
+
+                accumulator.wrapping_add(u64::from_be_bytes(
+                    (&bytes[x..y])
+                        .try_into()
+                        // SAFETY: OK, because `y - x = 8` and `wrapping_add` expects a `[u8; 8]`.
```
oconnor663, line 264: Just a terminology quibble: I wouldn't consider a fallible `unwrap`/`expect` to be a safety issue. A correctness issue for sure, but not something that can lead to UB.

src/test.rs
```diff
@@ -551,3 +551,31 @@ fn test_join_lengths() {
...
+        hasher.finish()
+    }
+
+    let hash = hash(crate::Hasher::new());
+    let expected = {
+        let mut hash = crate::Hasher::new();
+        hash.update(&2020u32.to_ne_bytes());
+        hash.update(&[7]);
+        hash.update(&[23]);
+        hash.update(&b"Hello, World!"[..]);
```
oconnor663, line 575: This is a good test, but I think it also highlights two of the reasons we've previously decided not to implement `Hasher`:

1. The `std::hash::Hash` serialization of standard types differs from machine to machine (`to_ne_bytes`).
2. I don't think there's any guarantee that the serialization of a type is stable over time. The standard library is within its rights to break this test in a future release.

oconnor663:
> Do you want Hasher to implement core::hash::Hasher?

As mentioned above, probably not. But let me know what you think of the discussion at #87.

> Do you find acceptable the strategy to convert Hash into u64?

Yes, I think there are three reasonable options:

1. modular addition, like you're doing here
2. xor
3. just truncating the hash to 8 bytes

The third option does seem kind of lazy compared to the first two, but it's what BLAKE3 already does if you request a short output.

> If OUT_LEN % 8 != 0, bytes will be missed by From<Hash> for u64. Can it be different? Do we have to assert that?

`OUT_LEN` is defined to be 32 bytes in the published spec, and that will never change. It's kind of like the values of the IV constants in that way. So I think it's fine to assume things like this in calling code.

src/lib.rs
Hywan: Correct!

src/lib.rs
```diff
@@ -239,6 +239,35 @@ impl From<Hash> for [u8; OUT_LEN] {
...
+        let bytes: [u8; OUT_LEN] = hash.into();
+
+        (0..(OUT_LEN / 8))
+            .into_iter()
+            .map(|i| (i * 8, (i + 1) * 8))
+            .fold(0u64, |accumulator, (x, y)| {
+                assert_eq!(y - x, 8);
+
+                accumulator.wrapping_add(u64::from_be_bytes(
+                    (&bytes[x..y])
```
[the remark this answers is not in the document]
Hywan: TIL, thanks!

Hywan:
Lot to read in order to formulate a counter point of view. Give me few days :-). Thanks for your feedback!

oconnor663:
I should add that I've been thinking a lot about "structured hashing" recently. (That term seems to be overloaded but I don't know what else to call it.) I feel like a widely useful standard could be doable. But at the same time, I don't think it would be a small project. Here's a random part of the question I was thinking about today, provoked by this PR:

---

How do we hash something like:

```rust
struct Foo {
    a: String,
    b: String,
}
```

Broadly speaking, it seems like there are two options. We either concatenate the bytes of the strings themselves with some kind of delimiter, or we hash them individually and concatenate their hashes. I'll call these approaches hashing together vs hashing separately. (For now I'll ignore mixed options like "hash them separately if they're above a certain length".) The upside of hashing separately is that the resulting structure is "more incremental" with respect to `b`. For example, say both `a` and `b` actually come from long streams of data, and we'd like to compute the final `Foo` hash incrementally without allocating a ton of memory to buffer either or both of the streams. In the hashing together approach, we'd start by feeding `a` into our final digest, we'd throw in some sort of counter and/or delimiter, and then after that we'd feed in `b`. This isn't ideal, because we can't make any progress reading the `b` stream until the `a` stream is entirely finished. That could be an issue in some interesting use cases.

If the programmer anticipated that this was going to be an issue, they could work around it by just defining `Foo` to hold the hashes of `a` and `b` rather than the strings themselves. That would effectively convert the hash together approach into the hash separately approach. However, there's a subtle security issue here, very reminiscent of the issues that come up in designing a tree hash. (`Foo` is kind of like a small tree with two leaves, right? When all I have is a hammer...) If we hash the strings separately, then "generalized length extension" attacks become possible. For example, if only know the hash of `a`, and not the contents of `a`, I can still compute the hash of a larger `Foo` struct that contains `a`. Again, this might be an issue in some interesting use cases. The defense again this sort of attack is to domain separate the hashes of interior fields from the "root" hash, like we do when we construct secure tree hashes, so that those interior hashes are never published. This isn't hard to do once the problem is clear, but it's something that one-off designs in the wild are likely to get wrong, and it's a way that a new standard might be able to distinguish itself.

---

Anyway, this is a big design space, and I probably shouldn't clutter up the BLAKE3 issues with too much discussion of it :) But if this is something that you or other folks want to discuss further, we could set something up.

PR #101, opened 2020-08-03
Fix cross compilation with clang-cl

Jake-Shadle:
When cross compiling with clang-cl for Windows, the build script was trying to use the Microsoft assembly files, which doesn't work with cc as it automatically assumes it can use the Microsoft assemblers. This just changes the logic to use the GAS files for Windows if we're targetting MSVC, but aren't actually on a Windows host.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -39,6 +39,23 @@ fn is_x86_64() -> bool {
     target_components()[0] == "x86_64"
 }
 
+fn is_windows_target() -> bool {
+    env::var("CARGO_CFG_TARGET_OS").unwrap() == "windows"
+}
+
+fn use_msvc_asm() -> bool {
+    // cc assumes if it is passed .asm and targetting MSVC that it can use the
+    // Microsoft assemblers, which isn't true when we're not on a Windows
+    // host, but are cross-compiling with clang-cl, so we explicitly check if
+    // we're on a Windows host (with the assumption they'll have the MSVC
+    // toolchain installed if they are, though that's not necessarily true)
+    if env::var("CARGO_CFG_TARGET_ENV").unwrap_or_default() == "gnu" {
+        false
+    } else {
+        env::var("HOST").unwrap_or_default().contains("-windows-")
+    }
+}
+
 fn is_x86_32() -> bool {
     let arch = &target_components()[0];
     arch == "i386" || arch == "i586" || arch == "i686"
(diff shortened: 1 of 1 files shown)


oconnor663:
Thanks for figuring this out. Can you suggest a way to add this build to [our CI testing](https://github.com/BLAKE3-team/BLAKE3/blob/master/.github/workflows/ci.yml)? This seems like the kind of thing that would be very easy to break under maintenance.

Jake-Shadle:
I agree it is a maintenance cost, unfortunately there is no convenient way to acquire the MSVCRT headers/libs on non-Windows hosts. That being said, I have fixed up similar code in other crates since cross-compiling to Windows is a pretty niche requirement, and I think some utilities could be added to cc-rs to make that use case more evident and make the logic cleaner in build scripts such as this one where there is both a GAS and MASM file for windows, and just let the cc-rs crate pick the appropriate one given the information that is has on the current toolset.

oconnor663:
@Sporif if you could come up with a branch that included CI testing, that would be awesome. If it ends up being too much trouble, though, let me know. Maybe I should've landed this a long time ago without worrying too much about testing.

messense:
> Unfortunately I can't get b3sum to build in CI (fails at lld-link linking) even though it builds locally.

Do you have any logs?

Sporif:
> > Unfortunately I can't get b3sum to build in CI (fails at lld-link linking) even though it builds locally.
> 
> Do you have any logs?

Yes, in this [action](https://github.com/Sporif/BLAKE3/runs/5376041760?check_suite_focus=true).

messense:
@Sporif Can you try run it with `-vv`?

Sporif:
> @Sporif Can you try run it with `-vv`?

Yes, here you go: https://github.com/Sporif/BLAKE3/runs/5403663935?check_suite_focus=true

messense:
Looks like it didn't pick up this `RUSTFLAGS`: https://github.com/messense/cargo-xwinbuild/blob/9f10a33f53be006c3b995df873a37bcc206f4d54/src/build.rs#L377-L382

messense:
@Sporif Can you try [cargo-xwinbuild v0.3.1-beta.1](https://crates.io/crates/cargo-xwinbuild/0.3.1-beta.1) ?

Sporif:
> @Sporif Can you try [cargo-xwinbuild v0.3.1-beta.1](https://crates.io/crates/cargo-xwinbuild/0.3.1-beta.1) ?

Yeah that fixes it, thanks!

https://github.com/Sporif/BLAKE3/runs/5404610375?check_suite_focus=true

messense:
@Sporif Thanks for confirming, v0.3.1 is released.

Sporif:
@oconnor663 Ok, so got all tests to pass except C bindings intrinsics which fails (default C bindings required fixing build.rs), and reference impl doc test which runs 0 tests for some reason.

Success: https://github.com/Sporif/BLAKE3/runs/5435012654?check_suite_focus=true
C intrinsics fail: https://github.com/Sporif/BLAKE3/runs/5434969357?check_suite_focus=true

PR #104, opened 2020-08-14
How to access the hex inside the Hash()

alexxroche:
blake3::hash(b"str") -> Hash(`hex`)  and I want the `hex` stripped out from the Hash() as a str

This change saves time so that the next person doesn't have to read the source code to locate .to_hex()


oconnor663:
Good idea, thank you. I'll land this, and then make my preferred changes on top of it.

PR #110, opened 2020-08-25
Add SSE2 implementations

mkrupcale:
### Introduction

This adds SSE2 intrinsic and assembly implementations based heavily on the respective SSE4.1 implementations. It simply emulates any SSE4.1 or SSSE3 intrinsics/instructions using SSE2 intrinsics/instructions. I'm certainly a novice in this area, so I can't claim that the implementations are optimal, but I hope that they are close, and the benchmarks certainly show an improvement over the portable versions.

### Details

There were two intrinsics which required emulation:

1. `_mm_blend_epi16` (SSE4.1): Used in both the rust and C intrinsic implementations
2. `_mm_shuffle_epi8` (SSSE3): Used in the C intrinsic implementation for `rot8` and `rot16`.

There were four instructions which required emulation:

1. `pblendw` (SSE4.1)
2. `blendvps` (SSE4.1)
3. `pinsrd` (SSE4.1)
4. `pshufb` (SSSE3)

Fortunately, these all had relatively simple SSE2 implementations.

### Testing

I've tested the SSE2 implementation on my own Linux (Fedora 32) machine (Phenom II X4) using the following methods:

1. `cargo test`
2. `cargo test --features 'prefer_intrinsics'`
3. `cargo bench`
4. `make -f Makefile.testing all`
5. `make -f Makefile.testing test`
6. `make -f Makefile.testing test_asm`

For the most part, things seem to work. Tests 1-3 above return some warnings like the following during compilation:

```sh
warning: unreachable statement
   --> src/platform.rs:408:5
    |
404 |           return true;
    |           ----------- any code following this expression is unreachable
...
408 | /     {
409 | |         if is_x86_feature_detected!("sse2") {
410 | |             return true;
411 | |         }
412 | |     }
    | |_____^ unreachable statement
    |
    = note: `#[warn(unreachable_code)]` on by default

warning: 1 warning emitted
```

I'm not sure if this is because SSE2 is always assumed to be present on x86_64 or if I'm missing something, but I essentially did the exact same thing as the SSE4.1 platform detection.

Test 4 fails to link with missing references to `g{,et}_cpu_features`:

```sh
$ make -f Makefile.testing all
gcc -O3 -Wall -Wextra -std=c11 -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2 -fPIE -fvisibility=hidden -Wa,--noexecstack -c blake3_sse2.c -o blake3_sse2.o -msse2
gcc -O3 -Wall -Wextra -std=c11 -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2 -fPIE -fvisibility=hidden -Wa,--noexecstack -c blake3_sse41.c -o blake3_sse41.o -msse4.1
gcc -O3 -Wall -Wextra -std=c11 -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2 -fPIE -fvisibility=hidden -Wa,--noexecstack -c blake3_avx2.c -o blake3_avx2.o -mavx2
gcc -O3 -Wall -Wextra -std=c11 -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2 -fPIE -fvisibility=hidden -Wa,--noexecstack -c blake3_avx512.c -o blake3_avx512.o -mavx512f -mavx512vl
gcc -O3 -Wall -Wextra -std=c11 -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2 -fPIE -fvisibility=hidden -Wa,--noexecstack blake3.c blake3_dispatch.c blake3_portable.c main.c blake3_sse2.o blake3_sse41.o blake3_avx2.o blake3_avx512.o -o blake3 -pie -Wl,-z,relro,-z,now
/usr/bin/ld: /tmp/cchzaxEt.o: in function `main':
main.c:(.text.startup+0x2ab): undefined reference to `get_cpu_features'
/usr/bin/ld: main.c:(.text.startup+0x2f5): undefined reference to `g_cpu_features'
collect2: error: ld returned 1 exit status
make: *** [Makefile.testing:46: all] Error 1
```

But this was already the case on the `master` branch on my machine.

Tests 5-6 seemed to work correctly.

I don't have a Windows machine to test the Windows GNU or MSVC assembly implementations, but they are nearly identical to the Unix implementation, so in principle they should work as well.

#### Benchmark results

I used the `BLAKE3-specs` criterion benchmarks to compare the `master` branch to my `sse2` branch[1]. For smaller inputs (`<= 1 kiB`), there were either statistically insignificant changes, no changes, or minor regressions of `~1-2%`. For medium inputs (`1 kiB < n < 10 kiB`), there were performance improvements ranging from `~2%` to `~200%`, eventually reaching `~230%` for the largest inputs.

[1] https://mkrupcale.gitlab.io/BLAKE3-specs-benchmark-results


sneves:
Supporting SSE2 was something I was not too sure about, given the vanishingly small percentage of machines still running without SSE4.1 (on the [Steam survey](https://store.steampowered.com/hwsurvey) 98.11% of machines have it). In any case, this is a solid patch.

I haven't gone over the whole thing, but one nit: the `_mm_blend_epi16` emulation routine is a bit complicated, especially if the compiler doesn't see past the expensive multiplication. I suggest rewriting it as
```
__m128i blend_epi16(__m128i a, __m128i b, const int imm8) {
  const __m128i bits = _mm_set_epi16(0x80, 0x40, 0x20, 0x10, 0x08, 0x04, 0x02, 0x01);
  __m128i mask = _mm_set1_epi16(imm8);
  mask = _mm_and_si128(mask, bits);
  mask = _mm_cmpeq_epi16(mask, bits);
  return _mm_or_si128(_mm_and_si128(mask, b), _mm_andnot_si128(mask, a));
}
```

The other nit is less technical: when emulating rotations with shift-shift-or, I find it preferable to use `por` instead of `pxor` to make it easily distinguished from the "real" xors in the algorithm. Though in fairness, we may have been not entirely consistent everywhere with that.

mkrupcale:
> Supporting SSE2 was something I was not too sure about, given the vanishingly small percentage of machines still running without SSE4.1 (on the [Steam survey](https://store.steampowered.com/hwsurvey) 98.11% of machines have it).

Yeah, I've seen the Steam survey as well, but an SSE2 baseline implementation would ensure 100% coverage of `x86_64` with optimized assembly. And incidentally my Phenom II falls into the category of CPUs without SSE4.1 :(.

> In any case, this is a solid patch.

Thanks, I'm glad I was able to produce something reasonable based on your work, having never before written assembly myself.

> I haven't gone over the whole thing, but one nit: the `_mm_blend_epi16` emulation routine is a bit complicated, especially if the compiler doesn't see past the expensive multiplication (the Microsoft compiler can't, from the looks of it). I suggest rewriting it as ...

Done in the latest commit.

> The other nit is less technical: when emulating rotations with shift-shift-or, I find it preferable to use `por` instead of `pxor` to make it easily distinguished from the "real" xors in the algorithm.

Makes sense.

> Though in fairness, we may have been not entirely consistent everywhere with that.

Perhaps this can be addressed in a separate cleanup PR covering all implementations.

sneves:
I've had some more time to review this patch. As far as the intrinsics changes go, I don't have any further comments. With respect to the assembly patches, I have some comments. Some of the tweaks below might also be applicable to the SSE4.1 code.

# blendvps patch

```diff
-        blendvps xmm1, xmm3, xmm0
-        blendvps xmm2, xmm4, xmm0
-        movdqa  xmmword ptr [rsp+0x110], xmm1
-        movdqa  xmmword ptr [rsp+0x120], xmm2
+        movdqa  xmm5, xmm0
+        movdqa  xmm6, xmm0
+        pandn   xmm5, xmm1
+        pand    xmm3, xmm6
+        por     xmm3, xmm5
+        movdqa  xmm5, xmm0
+        movdqa  xmm6, xmm0
+        pandn   xmm5, xmm2
+        pand    xmm4, xmm6
+        por     xmm4, xmm5
+        movdqa  xmmword ptr [rsp+0x110], xmm3
+        movdqa  xmmword ptr [rsp+0x120], xmm4
```

This operation arguably didn't need the blendvps instruction at all, and we can rewrite it better as

```diff
-        movdqa  xmm0, xmmword ptr [rsp+0x130]
-        movdqa  xmm1, xmmword ptr [rsp+0x110]
-        movdqa  xmm2, xmmword ptr [rsp+0x120]
-        movdqu  xmm3, xmmword ptr [rsp+0x118]
-        movdqu  xmm4, xmmword ptr [rsp+0x128]
-        blendvps xmm1, xmm3, xmm0
-        blendvps xmm2, xmm4, xmm0
-        movdqa  xmmword ptr [rsp+0x110], xmm1
-        movdqa  xmmword ptr [rsp+0x120], xmm2
+        mov     eax, dword ptr [rsp+0x130]
+        neg     eax
+        mov     r10d, dword ptr [rsp+0x110+8*rax]
+        mov     r11d, dword ptr [rsp+0x120+8*rax]
+        mov dword ptr [rsp+0x110], r10d
+        mov dword ptr [rsp+0x120], r11d
```

# pinsrd patch

```diff
-        pinsrd  xmm13, dword ptr [rsp+0x120], 1
-        pinsrd  xmm13, dword ptr [BLAKE3_BLOCK_LEN+rip], 2
+        mov     eax, dword ptr [rsp+0x120]
+        sar     eax, 16
+        pinsrw  xmm13, word ptr [rsp+0x120], 2
+        pinsrw  xmm13, eax, 3
+        mov     eax, dword ptr [BLAKE3_BLOCK_LEN+rip]
+        sar     eax, 16
+        pinsrw  xmm13, word ptr [BLAKE3_BLOCK_LEN+rip], 4
+        pinsrw  xmm13, eax, 5
...
-        pinsrd  xmm14, dword ptr [rsp+0x124], 1
-        pinsrd  xmm14, dword ptr [BLAKE3_BLOCK_LEN+rip], 2
+        mov     eax, dword ptr [BLAKE3_BLOCK_LEN+rip]
+        sar     eax, 16
+        pinsrw  xmm14, word ptr [rsp+0x124], 2
+        pinsrw  xmm14, eax, 3
+        mov     eax, dword ptr [BLAKE3_BLOCK_LEN+rip]
+        sar     eax, 16
+        pinsrw  xmm14, word ptr [BLAKE3_BLOCK_LEN+rip], 4
+        pinsrw  xmm14, eax, 5
...
-        pinsrd  xmm3, eax, 3
-        pinsrd  xmm11, eax, 3
+        mov     r14d, eax
+        sar     r14d, 16
+        pinsrw  xmm3, eax, 6
+        pinsrw  xmm3, r14d, 7
+        pinsrw  xmm11, eax, 6
+        pinsrw  xmm11, r14d, 7
```

There's a bug here where `mov     eax, dword ptr [BLAKE3_BLOCK_LEN+rip]` should be `mov     eax, dword ptr [rsp+0x124]`, but I suppose it's not caught by tests.

I would replace this by 

```diff
-        pinsrd  xmm13, dword ptr [rsp+0x120], 1
-        pinsrd  xmm13, dword ptr [BLAKE3_BLOCK_LEN+rip], 2
+        movd  xmm14, dword ptr [rsp+0x120]
+        punpckldq  xmm13, xmm14
...
-        pinsrd  xmm14, dword ptr [rsp+0x124], 1
-        pinsrd  xmm14, dword ptr [BLAKE3_BLOCK_LEN+rip], 2
+        movd  xmm13, dword ptr [rsp+0x124]
+        punpckldq  xmm14, xmm13
...
-        movaps  xmm3, xmmword ptr [rsp]
-        movaps  xmm11, xmmword ptr [rsp+0x10]
-        pinsrd  xmm3, eax, 3
-        pinsrd  xmm11, eax, 3
+        shl  rax, 0x20
+        or  rax, 0x40
+        movd  xmm3, rax
+        movdqa  xmmword ptr [rsp+0x20], xmm3
+        movaps  xmm3, xmmword ptr [rsp]
+        movaps  xmm11, xmmword ptr [rsp+0x10]
+        punpcklqdq  xmm3, xmmword ptr [rsp+0x20]
+        punpcklqdq  xmm11, xmmword ptr [rsp+0x20]

Similarly, we can simplify
```diff
-        pinsrd  xmm13, dword ptr [rsp+0x120], 1
-        pinsrd  xmm13, dword ptr [BLAKE3_BLOCK_LEN+rip], 2
+        mov     eax, dword ptr [rsp+0x120]
+        sar     eax, 16
+        pinsrw  xmm13, word ptr [rsp+0x120], 2
+        pinsrw  xmm13, eax, 3
+        mov     eax, dword ptr [BLAKE3_BLOCK_LEN+rip]
+        sar     eax, 16
+        pinsrw  xmm13, word ptr [BLAKE3_BLOCK_LEN+rip], 4
+        pinsrw  xmm13, eax, 5
```
to 
```diff
-        pinsrd  xmm13, dword ptr [rsp+0x120], 1
-        pinsrd  xmm13, dword ptr [BLAKE3_BLOCK_LEN+rip], 2
+        movd  xmm14, dword ptr [rsp+0x120]
+        punpcklqdq xmm13, xmm14
```
and
```diff
-        pinsrd  xmm3, eax, 3
+        mov     r14d, eax
+        sar     r14d, 16
+        pinsrw  xmm3, eax, 6
+        pinsrw  xmm3, r14d, 7
```
to
```diff
-        pinsrd  xmm3, eax, 3
+        shl    rax, 32
+        or     rax, 64
+        movd   xmm12, rax
+        movdqa xmm3, xmm13
+        punpcklqdq xmm3, xmm12
```
# pblendw patch

This one I don't have much quibbles with, but because in these older chips register moves still have a cost (they cost the same as a `pand(n)`, in fact), I feel it would be preferable to implement them as, for example,

```diff
-        pblendw xmm6, xmm5, 0xCC
+        pand    xmm6, xmmword ptr [PBLENDW_0x33_MASK+rip]
+        pand    xmm5, xmmword ptr [PBLENDW_0xCC_MASK+rip] # possibly a temp register here, if xmm5 is needed later
+        por     xmm6, xmm5
```

This forces fewer spills and moves, and the extra memory load isn't that bad, since we are not bound by memory uops in this loop.

For a concrete example,

```diff
-        pblendw xmm13, xmm12, 0xCC
+        movdqa  xmmword ptr [rsp+0x20], xmm2
+        movdqa  xmmword ptr [rsp+0x40], xmm3
+        movdqa  xmm2, xmmword ptr [PBLENDW_0xCC_MASK+rip]
+        movdqa  xmm3, xmm2
+        pandn   xmm2, xmm13
+        pand    xmm3, xmm12
+        movdqa  xmm13, xmm3
+        por     xmm13, xmm2
```

can become (we can also observe that `xmm12` is not used anymore after the `pblendw`, as `movdqa     xmm12,  xmm7` follows)

```diff
-        pblendw xmm13, xmm12, 0xCC
+        pand    xmm13, xmmword ptr [PBLENDW_0x33_MASK+rip] # -1, 0, -1, 0
+        pand    xmm12, xmmword ptr [PBLENDW_0xCC_MASK+rip] # 0, -1, 0, -1
+        por     xmm13, xmm12
```

which also lets us avoid the `xmm2, xmm3` spill.

mkrupcale:
I've made all the simplifications and changes suggested in the latest revision.

sneves:
Alright, I think we're good. Merging.

src/platform.rs
```diff
@@ -351,6 +391,20 @@ pub fn sse41_detected() -> bool {
...
+pub fn sse2_detected() -> bool {
+    // A testing-only short-circuit.
+    if cfg!(feature = "no_sse2") {
+        return false;
+    }
+    // Static check, e.g. for building with target-cpu=native.
+    #[cfg(target_feature = "sse2")]
+    {
+        return true;
+    }
```
oconnor663, line 405: The unconditional return was missing at the end here. Rustc almost always assumes that SSE2 is supported, but you can convince it not to and trigger a build break like this:

    cargo rustc --target i686-unknown-linux-musl -- -C target-cpu=i386

I've gone ahead and fixed this in https://github.com/BLAKE3-team/BLAKE3/pull/116.

oconnor663:
@mkrupcale thanks for doing so much work on this!

src/platform.rs
mkrupcale: > Rustc almost always assumes that SSE2 is supported, but you can convince it not to and trigger a build break like this:

Okay, that's what I suspected.

> I've gone ahead and fixed this in #116.

Thanks. I had something like that in my original version, but I wasn't sure how to get the tests to pass with the compiler warning that this part was unreachable.

PR #112, opened 2020-08-27
Implement `fmt::Debug` using builders

nvzqz:
This enables pretty printing via `{:#?}`. The normal style for `{:?}` is kept exactly the same.


oconnor663:
Awesome, TIL.

PR #114, opened 2020-08-30
C: Add blake3_hasher_init_derive_key_len

k0001:
`blake3_hasher_init_derive_key_len` is an alternative version of `blake3_hasher_init_derive_key` which takes the context and its length as separate parameters, and not together as a C string.

The motivation for this addition is making it easier for bindings to this C library to call this function without having to first copy over the context bytes just to add one `0x00` byte at the end.

Notice that contrary to `blake3_hasher_init_derive_key`, `blake3_hasher_init_derive_key_len` allows the inclusion of a `0x00` byte in the context. Given the rules about context string selection, this byte is unlikely to be used as part of a context string. But if for some reason it is ever given, it will be included in the context string and processed like any other non-alphanumeric byte would. For compatibility with `blake3_hasher_init_derive_key`, bindings should still check for the absence of `0x00` bytes. 

**See the relevant issue about the C string version**: #115

### DIFF
--- a/c/README.md
+++ b/c/README.md
@@ -103,30 +103,35 @@ Initialize a `blake3_hasher` in the keyed hashing mode. The key must be
 exactly 32 bytes.
 
 ```c
-void blake3_hasher_init_derive_key(
-  blake3_hasher *self,
-  const char *context);
-```
-
-Initialize a `blake3_hasher` in the key derivation mode. Key material
-should be given as input after initialization, using
-`blake3_hasher_update`. `context` is a standard null-terminated C string
-of any length.  The context string should be hardcoded, globally unique,
-and application-specific. A good default format for the context string
-is `"[application] [commit timestamp] [purpose]"`, e.g., `"example.com
-2019-12-25 16:18:03 session tokens v1"`.
-
-```c
-void blake3_hasher_init_derive_key_len(
+void blake3_hasher_init_derive_key_raw(
   blake3_hasher *self,
   const void *context,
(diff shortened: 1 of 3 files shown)


k0001:
Note that as described in #115, this is an intermediate solution. Ideally, I think `blake3_hasher_init_derive_key_len` should replace `blake3_hasher_init_derive_key` altogether.

k0001:
@oconnor663 I implemented the renaming you suggested in #115, and I added documentation to make things more clear.

@sneves I added a recommendation about the encoding of these bytes, but, since this is just a recommendation, I went ahead and recommend that the context string be written in a minimal ASCII alphabet. This is because many languages that allow you to write unicode literals directly don't necessarily represent this as UTF-8, but as some unicode abstraction, and moreover, because encoding unicode string literals to UTF-8 necessarily requires a unicode normalization step if you are expecting to use said UTF-8 bytes as cryptographic material. Of course, users are free to use UTF-8 if they want, but to keep things very simple and error-proof, a basic ASCII alphabet is what we should be recommending.

Here's the rendered version of the README: https://github.com/BLAKE3-team/BLAKE3/blob/b205e0efa1a1390dd91e3c015976e3d4ce999fff/c/README.md

oconnor663:
Thanks for this, and apologies for the delay. I'll add some edits of my own to the C readme on top of this, and maybe some extra tests.

k0001:
Great! Thanks for taking the time to consider this change!

PR #119, opened 2020-09-29
fix the C implementation on big endian, and add CI coverage for that case

oconnor663:
Previously we were covering the Rust implementation on big endian, but not C. With this change their testing stories will be pretty similar. (There's plenty of duplicated code in `blake3_c_rust_bindings` that I'm not excited to maintain, not to mention another hacky `cross_test.sh` script, but it works.)


oconnor663:
This fix will merit a new "release" for our C code, which at this point is just a `git tag`. Will make a note to self to do that within the next couple days.

Issue #132, opened 2020-11-02
Turing the `b3sum` crate into a library (with an executable)


bhgomes:
I'm planning on using BLAKE3 in a project but really only want to use the `b3sum` functionality. Would it be possible to turn the `b3sum` crate into a library (with an executable) so that I can use say the `write_hex_output` function instead of having a copy of it in my library? Right now, I am relying on the `b3sum` executable to be installed and running a new process every time I want to use it.

oconnor663:
I'm not necessarily opposed, though if it was just for `write_hex_output` I'm not sure it would be worth it. I think the real meat of `b3sum` is in the memory mapping and multithreading that it does. It's not nearly as complicated as what's in `blake3` itself, but it's still a little much for folks who haven't dealt with memory mapping before, especially because it necessarily involves unsafe code. Exposing a function like `b3sum::hash_this_file_the_fastest_way_you_can_including_maybe_multithreading()` might be a win.

bhgomes:
Right, I meant that function as a simple use case for my application, but whatever makes sense for the user base is great.

oconnor663:
Well so far the user base is...pretty much you :) Or at least, you're the only one who's asked us about it so far. Do you think you'd have time to put together a PR for this? I'd be happy to review it and help you get it landed.

bhgomes:
Yeah, I'll try to put together the first draft PR by this weekend. Thanks.

PR #133, opened 2020-11-08
Extract `b3sum` functionality to a library file

bhgomes:
See #132.

## PR Changes
- Use `HasherGenerator` to remember key strategies and avoid the stdin/keyed conflict
- Add functions that hash output to arbitrary writers, not just stdout.
- Add `CheckLine` structure which holds a parsed checkline and can be checked independently

### `Mmap` Interface

Right now, all functions pass the `no_mmap` argument. It would be nice to have an interface that automatically sets `no_mmap` to `false` and then if the user specifically wants `no_mmap` then they can use another function say using some `no_mmap` module or `possibly_no_mmap` module.

### Checkfile Checking Interface

The library has at least three ways of checking a checkfile `check_checkfile`, `check_checkfile_iter`, and `print_check_checkfile`, and none are ideal. Code is repeated amongst all three and `check_checkfile_iter` is probably not the correct implementation, since the errors encountered in reading lines from `BufReader` are emitted inside the iterator and not in the enveloping `Result` type.

### Multithreading

In the `main` function we are using the same multithreading strategy as before the PR, but the library functions are not taking advantage of the threading since we don't necessarily install into a thread pool. Not sure what is the best interface for this. Perhaps we have the fast defaults threading+mmap and then another module has the slow versions.

### DIFF
--- a/b3sum/src/lib.rs
+++ b/b3sum/src/lib.rs
@@ -1,0 +1,599 @@
+use anyhow::{bail, ensure, Error, Result};
+use std::cmp;
+use std::fs::File;
+use std::io;
+use std::io::prelude::*;
+use std::iter::from_fn;
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
+
+#[cfg(test)]
+mod unit_tests;
+
+pub const NAME: &str = "b3sum";
+
+/// Key Strategy
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum KeyStrategy {
+    Keyed,
+    DeriveKey,
+    Default,
+}
+
+/// Hasher which remembers its `KeyStrategy`
+#[derive(Clone, Debug)]
+pub struct HasherGenerator {
+    hasher: blake3::Hasher,
+    key_type: KeyStrategy,
+}
+
+impl HasherGenerator {
+    /// Make a `HasherGenerator` which produces keyed hashers.
+    pub fn new_keyed(key: &[u8; blake3::KEY_LEN]) -> Self {
+        Self {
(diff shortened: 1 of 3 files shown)


oconnor663 (commented):
Thanks for working on this!

> The library has at least three ways of checking a checkfile check_checkfile, check_checkfile_iter, and print_check_checkfile, and none are ideal.

Yes, this version might be adding more complexity than we want. Maybe I'm about to get too abstract :) but in my mind there's kind of a no-man's-land in between "simple functions on compartmentalized data types" and "application code that assumes it's the main thread and feels free to parse argv and print to stdout." I think the iterator code might be sitting in that no-man's-land.

If I were going to represent `--check` in simple data types, my first instinct would be to provide an iterator that parses the checkfile, and to leave the actual hashing (provided as a separate function) and comparing to the caller's own `for` loop. (This leaves open the possibility that the caller might want to use multiple threads to check files in parallel, for example.) But actually before that, I'd want to take a larger step back and figure out whether we really need it in the first place. For example in your use case, do you need careful control over exactly how `--check` works? Do you even need `--check`? Could we get away with a simpler first step that, say, only exposes hashing functions?

b3sum/src/lib.rs
```diff
@@ -0,0 +1,599 @@
...
+use std::io;
+use std::io::prelude::*;
+use std::iter::from_fn;
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
+
+#[cfg(test)]
+mod unit_tests;
+
+pub const NAME: &str = "b3sum";
```
Rudxain, line 13: I think `NAME` belongs to `main.rs`, it doesn't make much sense to be part of `lib`

PR #135, opened 2020-11-28
README: derive_key should be used with high entropy keys

ericonr:
The Rust example can be understood as a recommendation to use the
derive_key function as a PKDF, since the example input_key can be
interpreted as such. As a KDF, derive_key requires a high entropy
input_key, and that should be made clear.


oconnor663:
Yes, and the readme does include a "BLAKE3 is not a password hashing algorithm" paragraph above. Can you suggest a way to improve this example without making it too complicated? We could use `rand` to generate some random bytes, but then maybe it wouldn't be sufficiently clear to readers that `rand::random` is cryptographically secure (and that that's a requirement)?

ericonr:
Do you know a simple rust code snippet that would clearly be equivalent to reading some bytes from `/dev/urandom`? That might be a good fit, I think.

Or even a comment in the code...

```rust
// cryptographically secure random buffer
let random_buffer = rand::random(...);
```

PR #141, opened 2021-01-05
Added optimized scala implementation

catap:


oconnor663:
That section is getting a little crowded, and it's not meant to be exhaustive. We do try to tweet each new implementaiton we find out about, though, and I've just tweeted this one :) https://twitter.com/Blake3Team/status/1346485887236853760

Out of curiosity, what would be the target for a Scala implementation as opposed to the existing JNI bindings? Is it preferable in cases where you don't necessarily have a working C compiler?

catap:
@oconnor663 this implementation mainly focused to support [Scala.JS](https://www.scala-js.org) and [Scala Native](http://scala-native.org/).

JNI bindings aren't available at this two platform.

PR #147, opened 2021-02-03
Draft: serde support

maisiliym:
- Implement optional ser/deserialize attribute for `Hash`
- serde?/std weak dependency reference: https://github.com/rust-lang/cargo/issues/8832

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,6 +78,7 @@ rayon = { version = "1.2.1", optional = true }
 cfg-if = "0.1.10"
 digest = "0.9.0"
 crypto-mac = "0.8.0"
+serde = { features = ["derive"], version = "1.0.123", optional = true }
 
 [dev-dependencies]
 hex = "0.4.2"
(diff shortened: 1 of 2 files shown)


oconnor663:
What are the tradeoffs with implementing Serde traits on `Hash`, vs having the caller convert a hash to `[u8; 32]` with `.into()` and serialize those bytes directly? I tend to think of `Hash` as a convenience wrapper to make `.to_hex()` available, and as an opportunity for the documentation to ask the user to think about constant-time comparisons. But callers who plan on reading and writing the bytes of a hash will probably have an easier time if they convert the hash to bytes immediately?

oconnor663:
> serde?/std weak dependency reference: rust-lang/cargo#8832

That's interesting. Does it work on stable?

sshine:
The use-case I have is a program that parameterises the hash digest/function. Here it is convenient to write trait instances for `blake3::Hash` and `blake3::Hasher`. If I were to operate directly on `[u8; _]`, I couldn't differentiate between different hash digests/functions that have an inner byte-array representation. In my case, no matter what hash digest/function I'm choosing, I need to be able to serialize/deserialize the digests. If I could write

```toml
[dependencies]
blake3 = { version = "1.2", features = ["serde"] }
```

and get those instances, that would save me from creating my own alternative wrapper to `blake3::Hash`.

jsenkpiel-godaddy:
This would be very nice for having my struct be able to serialize/deserialize directly without an intermediate.

Right now I have to have a `B3Hash` newtype wrap `blake3::Hash` with manual `Serialize`, `Deserialize`, `Visitor`, etc.

PR #153, opened 2021-02-19
add OutputReader::xor() which XOR the buffer with the data

PaulGrandperrin:
can be useful to construct a stream cipher

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1356,6 +1356,26 @@ impl OutputReader {
         }
     }
 
+    /// Similar to [`fill`] but instead of overriding the buffer this method XOR the data.
+    ///
+    /// [`fill`]: #method.fill
+    pub fn xor(&mut self, mut buf: &mut [u8]) {
+        while !buf.is_empty() {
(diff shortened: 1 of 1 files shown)


oconnor663:
We've thought about the stream cipher use case a few times, though we haven't gotten around to implementing it yet. The biggest missing piece is that we need the output phase to use the same SIMD optimizations that the input phase uses, which will require some substantial internal refactoring. It might be that we want to move the xor operations all the way down into SIMD/assembly, to take advantage of vector ops like `_mm256_xor_si256`. (Then again, maybe the efficiency gain there would be too small to measure and not worth the complexity.)

Another interesting option here, is that an operation like `.xor()` can be _multithreaded_. For example, say you have a large file on disk, and you want to encrypt it with this stream cipher. The application could memory map the file, and then we could use `rayon` to divide-and-conquer the whole thing. This is basically what `b3sum` currently does for input. (On the other hand, encrypting large files usually means we want to insert MAC tokens into the output, so a giant xor-in-place might not be as useful.)

Anyway, I'm thinking we might want to try to implement some of this stuff, before we commit to a particular public API?

PR #154, opened 2021-02-19
add OutputReader::fill_uninit() which accepts a `&mut [MaybeInit<u8>]…

PaulGrandperrin:
to avoid unnecessary memory initialization


oconnor663:
Is there a safe conversion between something like `&mut MaybeUninit<[u8; 64]>` and `&mut [MaybeUninit<u8>]`? It seems like there could be, along the lines of conversions that exist for `Cell`, but I haven't seen one yet for `MaybeUninit`. What would a caller look like?

Issue #122, opened 2020-10-09
Backwards incompatible changes in Blake3 0.3.5


asomers-ax:
Blake3 0.3.5 updated Digest from 0.8.1 to 0.9.0.  That brought in several backwards-incompatible changes, as Digest's CHANGELOG shows.  Changes like this should only be made with a minor version bump (or major version if you're past 1.0.0).

https://github.com/RustCrypto/traits/blob/master/digest/CHANGELOG.md

oconnor663:
There was some discussion here: https://github.com/BLAKE3-team/BLAKE3/pull/90

I have mixed feelings about this. Of course you're correct, that anything that's likely to break calling code should strictly speaking be indicated as a breaking change in the version number. On the other hand, I've been on the fence about including `digest` trait implementations in this crate, for exactly that reason. This crate is generally pretty stable (the pre-1.0 version number notwithstanding), so doing an incompatible version bump because of a version bump in what is effectively an optional dependency feels like the tail wagging the dog to me. A "better" solution would be to have a third crate, which implements `digest` traits on thin wrappers around `blake3` types. However, that would be more crate names to keep in our heads, and lots of folks would get confused by compiler errors about how `blake3::Hasher` doesn't implement `digest::Digest`. ("You can't use `blake3::Hasher` here, you need to use `blake3_digest::Hasher`." Literally unplayable.)

In the end, I've kind of settled on this "dirty" approach, where in practice (one time so far) I've treated the `digest` dependency as exempt from the stability guarantee of this crate. I assuage my own guilt over this by focusing on how in the distant future, when both crates are 1.0, thing will be stabler. If it looks like this is going to keep happening with any regularity, I should explicitly document it, but it's not clear to me that it will.

PR #157, opened 2021-02-25
Cargo.toml: update crypto-mac to 0.10.0

PaulGrandperrin:
This is an API breaking change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,7 +77,7 @@ constant_time_eq = "0.1.5"
 rayon = { version = "1.2.1", optional = true }
 cfg-if = "0.1.10"
 digest = "0.9.0"
-crypto-mac = "0.8.0"
+crypto-mac = "0.10.0"
 
 [dev-dependencies]
 hex = "0.4.2"


oconnor663:
I've gone back and forth in the past about how properly to update `digest` and `crypto-mac` in the past:

- https://github.com/BLAKE3-team/BLAKE3/pull/90
- https://github.com/BLAKE3-team/BLAKE3/issues/122

We're thinking about tagging this crate 1.0 any day now, so this is going to become more important. I don't think anyone wants to bump the crate to 2.0 when `crypto-mac` gets to 0.11. What's the right arrangement here? @tarcieri?

tarcieri:
@oconnor663 in the `signature` crate (which is post-1.0) we've opted to have a set of "unstable features" with `*-preview` names which depend on the pre-1.0 RustCrypto trait crates (e.g. `digest`):

https://docs.rs/signature/1.3.0/signature/#unstable-features

We explicitly call these out-of-scope for semver, but have a policy that we bump the *minor* version when the trait crates are updated.

It's sort of similar to having a 0.x crate for these features, but the other parts of the API are covered under a "1.0" semver guarantee.

tarcieri:
Also note: we'll be releasing `crypto-mac` to v0.11 sometime in the next few weeks.  See: 

https://github.com/RustCrypto/traits/issues/392

oconnor663:
The follow-up commit 07b746b1b4792f8885fe8749b22dd6242467b674 implements @tarcieri's suggestion, similar to the `digest` crate. (Though I don't think we'll make the nuanced guarantee about patch vs minor version bumps.)

tarcieri:
Hopefully we'll be able to get to 1.0 sometime this year. Depends how quickly const generics evolve (we don't want to ship a 1.0 with `generic-array`)

PR #161, opened 2021-03-01
Proposal: privatize the Join trait and expose *_rayon methods

oconnor663:
The `Join` trait is 1) pretty difficult to understand in the first place, 2) pretty unlikely to ever be used by anyone directly, and 3) a pretty big exposure of implementation details. I don't think we want to stabilize it in 1.0. My current proposal (this PR) is to un-`pub` it and to instead expose a collection of `*_rayon` functions: `hash_rayon()`, `Hasher::update_rayon()`, etc.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,10 +25,9 @@ neon = []
 # entire build, with e.g. RUSTFLAGS="-C target-cpu=native".
 std = ["digest/std"]
 
-# The "rayon" feature (defined below as an optional dependency) enables the
-# join::RayonJoin type, which can be used with Hasher::update_with_join to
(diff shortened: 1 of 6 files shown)


oconnor663:
Thanks! This seems like a good compromise.

In some other projects I've [added `src/bin/` executables](https://github.com/oconnor663/duct.rs/tree/master/src/bin) to take care of this in a maximally portable way. The main downside there is that now the library has binary build products, which is kind of weird. I guess the cleanest possible solution would be to put these targets in a separate crate, and we can do that if it ends up solving a real world problem for someone, but I've never gotten around to it.

Let me know if you need this change released on crates.io?

PR #164, opened 2021-03-22
Update build.rs

goodengineer:


build.rs
```diff
@@ -156,7 +156,7 @@ fn build_sse2_sse41_avx2_assembly() {
 
 fn build_avx512_c_intrinsics() {
     // This is required on 32-bit x86 targets, since the assembly
-    // implementation doesn't support support those.
+    // implementation doesn't support  those.
```
oconnor663, line 159: Could you remove the double space after `support`?

oconnor663:
Thanks for the corrections. It's funny, even looking _right at your change_ it still took me a minute to figure out what was wrong with the original. Like my brain just would not see it.

PR #165, opened 2021-03-22
comment cleanup in lib.rs

goodengineer:


src/lib.rs
```diff
@@ -72,7 +72,7 @@
 #[cfg(test)]
 mod test;
 
-// The guts module is for incremental use cases like the `bao` crate that need
```
oconnor663, line 75: I think "need" is actually correct here, since the subject of the clause is "cases".

Issue #168, opened 2021-04-28
Please document collision-resistance properties of BLAKE3 with less than 256 bits


joshtriplett:
I'm interested in the collision-resistance and safety of BLAKE3 with less than 256 bits (either via XOF or truncation), for use in applications storing a very large number of hash values and sensitive to storage size.

Would BLAKE3 with 224 bits be considered as safe as SHA-224, SHA-512/224, or SHA3-224? Is that likely to be reasonably safe for years to come?

Would any shorter number of bits be acceptably secure?

If so, what would be the preferred method of obtaining such hashes? XOF, or truncation?

Would those answers vary for other BLAKE-family hashes, which use a larger number of rounds?

(I entirely understand that hash length is a tradeoff. I'm trying to find an acceptable balance between degree of future-proofing and usability in storage-size-sensitive applications.)

oconnor663:
Btw, if you happen to be working on tracking build artifacts, that's right up my alley. I was on the Buck team at Facebook back in the day, and a friend and I still maintain a tool called [`peru`](https://github.com/buildinspace/peru) that's kind of in the same space.

sneves:
Any better-than-generic attack that works for n-t-bit output will also apply to the full n-bit output, as you're only giving the attacker more information to work with in that case.

joshtriplett:
Thanks for the explanation!

Would you consider documenting those details somewhere in the BLAKE3 documentation? Most notably, that BLAKE3 at N bits is collision-resistant to N/2 bits, and that XOF and truncation are equivalent?

oconnor663:
@joshtriplett I just opened https://github.com/BLAKE3-team/BLAKE3/pull/169 with some proposed doc changes. Let me know what you think.

joshtriplett:
That looks great! Thank you very much.

oconnor663:
Merged.

PR #169, opened 2021-04-30
explicitly document the properties of short outputs

oconnor663:
Suggested by @joshtriplett at:
https://github.com/BLAKE3-team/BLAKE3/issues/168#issuecomment-829609667

@sneves what do you think of the wording here?

### DIFF
--- a/c/README.md
+++ b/c/README.md
@@ -92,10 +92,21 @@ void blake3_hasher_finalize(
   size_t out_len);
 ```
 
-Finalize the hasher and emit an output of any length. This doesn't
-modify the hasher itself, and it's possible to finalize again after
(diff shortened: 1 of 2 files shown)


sneves:
That seems fine, however I don't think we've claimed specific values for collision and preimage resistance in the past, simply a generic min(outlen/2, 128)-bit security claim.

oconnor663:
Using keyed BLAKE3 to generate a 16-byte MAC seems like a pretty reasonable use case to me, though, and I suppose technically we haven't given users permission to do that until we say "preimage reiststance is min(outlen, 256)"?

oconnor663:
Ok, I'm going to land this as is, but we might come back and change the wording at some point regarding security claims > 128 bits.

PR #173, opened 2021-05-18
Release 1.0

oconnor663:
```
Changes since 0.3.7:
- Fix a build break under Visual Studio 2015:
  https://github.com/BLAKE3-team/BLAKE3/issues/142
- Add Hash::from_hex().
- Implement PartialEq<[u8]> for Hash, using constant_time_eq.
- Implement Display for Hash, equivalent to Hash::to_hex().
- Change derive_key() to return a 32-byte array. As with hash() and
  keyed_hash(), callers who want a non-default output length can use
  Hasher::finalize_xof().
- Replace Hasher::update_with_join() with Hasher::update_rayon(). The
  former was excessively generic, and the Join trait leaked
  implementation details. As part of this change, the Join trait is no
  longer public.
- Upgraded arrayvec to 0.7.0, which uses const generics. This bumps the
  minimum supported Rust compiler version to 1.51.
- Gate the digest and crypto-mac trait implementations behind an
  unstable feature, "traits-preview". As part of this change upgrade
  crypto-mac to 0.11.0.
```

The upgrade to arrayvec 0.7.0 involves a bunch of code changes, replacing the old hacky `arrayvec::Array` trait with nice new const generic paramters across a bunch of internal functions. But this is all just mechanical Rust type system stuff, no behavioral change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "blake3"
-version = "0.3.7"
+version = "1.0.0"
 authors = ["Jack O'Connor <oconnor663@gmail.com>"]
 description = "the BLAKE3 hash function"
 repository = "https://github.com/BLAKE3-team/BLAKE3"
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ output_reader.fill(&mut output);
 assert_eq!(&output[..32], hash1.as_bytes());
 
 // Print a hash as hex.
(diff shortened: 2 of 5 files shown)


oconnor663:
(Note that 0.3.8, a backport of bugfixes from master, was just released earlier today. For anyone following along, those commits will go out "again" as part of this 1.0 release.)

PR #175, opened 2021-05-24
Compute threaded checksums of chunks, iteratively

### LINKED ISSUE #31: Poor speed
When using blake2b I get around 13sec when hashing 941MB file using *blake2_bin* and *rash* crates but I get about 30sec when hashing the same file using *b3sum* crate. The environment is rust 1.40 on Ubuntu 18.04 using Core i3

lefth:
This seems to improve performance slightly on SSDs, but more than doubles
performance on spinning hard disks.

The buffer size seems to make a noticeable difference. I found 4 megabytes works well on my computer, but it may depend on the number of CPU cores available. Advice would be appreciated unless this value is likely to work well in most cases.

Fixes #31.

### DIFF
--- a/b3sum/src/main.rs
+++ b/b3sum/src/main.rs
@@ -219,7 +219,9 @@ impl Input {
             // multiple threads. This doesn't work on stdin, or on some files,
             // and it can also be disabled with --no-mmap.
             Self::Mmap(cursor) => {
-                hasher.update_rayon(cursor.get_ref());
+                for chunk in cursor.get_ref().chunks(4_194_304) {
+                    hasher.update_rayon(chunk);
+                }
             }
             // The slower paths, for stdin or files we didn't/couldn't mmap.
             // This is currently all single-threaded. Doing multi-threaded


lefth:
It looks like the test server failed to install Rust in one case, rather than any test code failing.

https://github.com/BLAKE3-team/BLAKE3/pull/175/checks?check_run_id=2656464255

oconnor663:
> it may depend on the number of CPU cores available

Yes it certainly does. A big AWS machine with ~40 cores is going to need much larger chunks to get good throughput.

At the same time, while this mitigates the spinning disk issues, it doesn't really _fix_ them. It just means you thrash over a smaller chunk of the file, rather than thrashing over the whole thing. The real fix for this -- besides just disabling threading entirely -- is the "from the front" approach described in #31. I have played around with this a few different times in the last couple years, most recently with a couple small commits on [this branch](https://github.com/BLAKE3-team/BLAKE3/tree/offload), but I haven't yet put in the time to make something high-quality. I think this approach will ultimately become our default. We'll accept a ~10% performance hit in common cases, but the thrashing issue on spinning disks will go away, and we'll get a big performance improvement in piping use cases like `cat file | b3sum`.

lefth:
@oconnor663 As far as I understand, a "from the front" approach would mean making `update_with_join` spawn threads that read sequential sections rather than dividing up the file. To contrast with this PR, the "from the front" chunk would be a moving window instead of explicit reads (still using mmap for the whole file by default, or a `Read` trait for stdin). The OS would choose the actual read sizes... I wonder if that could be done with no performance penalty at all, since that change is like a better version of this PR (which itself has no performance penalty with 8 cores).

Are there issues I'm not guessing, perhaps overhead due to too small computations?

oconnor663:
In my mind, `update_with_join` (which is being simplified and renamed to `update_rayon` in the [upcoming 1.0 release](https://github.com/BLAKE3-team/BLAKE3/pull/173)) doesn't need to change. The recursive-divide-and-conquer-work-stealing approach it uses today is probably the most generally efficient way to hash a big memory buffer, and "use Rayon to hash a big memory buffer as efficiently as possible" is that function's job. The performance issues that come up with `b3sum` stem from the fact that what appears to be memory buffer (`&[u8]`) is actually a magical mmap thing with (sometimes) pathological thrashing behavior. I think right response here is to define a new interface for this case that takes different arguments and makes different guarantees, and not to try to optimize for two very different cases in a single interface.

The best way to express that new interface in Rust is probably the `Read` trait. In particular, `File` and `Stdin` both implement `Read`, as does `TcpStream`. I imagine `blake3::Hasher` could add a method called something like `update_from_reader(r: impl Read) -> io::Result<u64>`. (Note that `Read` has a blanket impl that covers `&mut impl Read`.) Inside, the calling thread would read its argument into large buffers (somewhere between 16 KiB and 1 MiB, we'd need to benchmark and tune), send those to Rayon tasks, and receive subtree chaining values or similar from them through channels.

oconnor663:
By the way, if you or anyone else reading along would be interested in working on a `Read`-based implementation, let me know. I'd be happy talk in depth about what such an implementation would look like, mentor a PR, etc. (Otherwise I still plan on getting to it "soon"/"eventually".) There are some subtleties to do with how subtree chaining values get represented (if you hash e.g. 1 MiB at a time, you have to worry about getting the right answer for an input that's exactly 1 MiB long), but I think I know the Right Way to handle those subtleties (namely, pass around the internal `Output` type instead of `[u8; 32]` chaining values). The real meat & potatoes of buffering, queueing, and threading don't depend too much\* on how BLAKE3 works on the inside once that detail is taken care of.

\* There's another subtlety to do with "what if we've already input say 1 KiB of input into the `Hasher`", but we could sidestep that at first by just asserting that the `Hasher` is fresh. In practice, it would almost always be fresh, and the unusual cases can be dealt with later, once we've got something working.

PR #185, opened 2021-07-26
`cargo clippy` to fix Rust code style guidelines

lrazovic:
* Using the `include` field in the `Cargo.toml` the crate is now 138Kb vs 155kB.
* Run `cargo clippy` to fix Rust styles:
  * Add the `is_empty()` method to the `ChunkState` struct. [Clippy warning](https://rust-lang.github.io/rust-clippy/master/index.html#len_without_is_empty)
  * Remove `Hash` derivation since `PartialEq` is explicitly implemented. [Clippy warning](https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq)
  * Use `''` instead of `""` in `contains()` and `split()` [Clippy warning](https://rust-lang.github.io/rust-clippy/master/index.html#single_char_pattern)
  * Remove `Hash` derivation in the `Hash` struct since `PartialEq` is implemented [Clippy warning](https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq)
  * Remove a needless borrow since the `key` is going to be dereferenced immediately by the compiler [Clippy warning](https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ license = "CC0-1.0 OR Apache-2.0"
 documentation = "https://docs.rs/blake3"
 readme = "README.md"
 edition = "2018"
+include = ["src/**/*", "LICENSE", "README.md", "c/*", "build.rs"]
 
 [features]
 default = ["std"]
--- a/build.rs
+++ b/build.rs
@@ -22,7 +22,7 @@ fn is_ci() -> bool {
 }
 
 fn warn(warning: &str) {
-    assert!(!warning.contains("\n"));
+    assert!(!warning.contains('\n'));
     println!("cargo:warning={}", warning);
     if is_ci() {
         println!("cargo:warning=Warnings in CI are treated as errors. Build failed.");
@@ -32,7 +32,7 @@ fn warn(warning: &str) {
 
 fn target_components() -> Vec<String> {
     let target = env::var("TARGET").unwrap();
-    target.split("-").map(|s| s.to_string()).collect()
+    target.split('-').map(|s| s.to_string()).collect()
 }
 
 fn is_x86_64() -> bool {
--- a/src/guts.rs
+++ b/src/guts.rs
@@ -30,6 +30,11 @@ impl ChunkState {
     }
 
     #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    #[inline]
     pub fn update(&mut self, input: &[u8]) -> &mut Self {
         self.0.update(input);
         self
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -437,6 +437,10 @@ impl ChunkState {
         BLOCK_LEN * self.blocks_compressed as usize + self.buf_len as usize
     }
 
+    fn is_empty(&self) -> bool {
+        self.blocks_compressed + self.buf_len == 0
+    }
+
     fn fill_buf(&mut self, input: &mut &[u8]) {
         let want = BLOCK_LEN - self.buf_len as usize;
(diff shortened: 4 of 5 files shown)


src/lib.rs
```diff
@@ -198,7 +198,7 @@ fn counter_high(counter: u64) -> u32 {
 /// [`from_hex`]: #method.from_hex
 /// [`Display`]: https://doc.rust-lang.org/std/fmt/trait.Display.html
 /// [`FromStr`]: https://doc.rust-lang.org/std/str/trait.FromStr.html
-#[derive(Clone, Copy, Hash)]
+#[derive(Clone, Copy)]
```
oconnor663, line 201: I understand the general intent of this lint, but in this case I think `derive(Hash)` is fine. Any two BLAKE3 hashes that compare equal will also give the same HashMap hash. (The current `PartialEq` implementation has the same behavior that `derive(PartialEq)` would, with the important difference being that it's done in constant time.) More importantly, removing `derive(Hash)` here without an explicit impl to replace it would be a breaking change.

src/guts.rs
```diff
@@ -29,6 +29,11 @@ impl ChunkState {
         self.0.len()
     }
 
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
```
oconnor663, line 35: This type is almost entirely internal. (Technically public, but hidden and unstable.) I'd like to avoid expanding it with convenience methods for now.

build.rs
```diff
@@ -22,7 +22,7 @@ fn is_ci() -> bool {
 }
 
 fn warn(warning: &str) {
-    assert!(!warning.contains("\n"));
+    assert!(!warning.contains('\n'));
```
oconnor663, line 25: :+1:

src/lib.rs
```diff
@@ -760,7 +764,7 @@ fn compress_subtree_to_parent_node<J: join::Join>(
     debug_assert!(input.len() > CHUNK_LEN);
     let mut cv_array = [0; MAX_SIMD_DEGREE_OR_2 * OUT_LEN];
     let mut num_cvs =
-        compress_subtree_wide::<J>(input, &key, chunk_counter, flags, platform, &mut cv_array);
+        compress_subtree_wide::<J>(input, key, chunk_counter, flags, platform, &mut cv_array);
```
oconnor663, line 767: :+1:

oconnor663 (changes requested):

oconnor663:
Thanks for the cleanup. My past experiences with Clippy are that it has a lot of false positives in this crate. (Or I should say, things that I would personally prefer not to change 😄 ) But things like avoiding unnecessary `&`'s are good fixes, let's definitely do those.

lrazovic:
Thank you so much for the comprehensive comments, very helpful and well done. I should have made the requested changes, I hope they meet the specifications.

Cargo.toml
```diff
@@ -8,6 +8,7 @@ license = "CC0-1.0 OR Apache-2.0"
 documentation = "https://docs.rs/blake3"
 readme = "README.md"
 edition = "2018"
+include = ["src/**/*", "LICENSE", "README.md", "c/*", "build.rs"]
```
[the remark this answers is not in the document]
lrazovic: I have now excluded all folders that are not needed for the use of the crate as external library.

lrazovic:
Congrats again on the result and implementation, really good work. If I may say so, maybe you could specify in the `CONTRIBUTING.md` that the code has already been checked by `cargo clippy`, in order to avoid not very useful PR like this one.

PR #188, opened 2021-07-30
explicitly #error on big-endian ARM

oconnor663:
This check was suggested by `brandmeyer` on HN:

https://news.ycombinator.com/item?id=27979437

Before landing this, I need to figure out how to test it, ideally in CI.

### DIFF
--- a/c/blake3_neon.c
+++ b/c/blake3_neon.c
@@ -2,7 +2,12 @@
 
 #include <arm_neon.h>
 
-// TODO: This is probably incorrect for big-endian ARM. How should that work?
+#ifdef __ARM_BIG_ENDIAN
+#error "This implementation only supports little-endian ARM."
+// It might be that all we need for big-endian support here is to get the loads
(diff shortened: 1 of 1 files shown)


jbrandmeyer:
Thanks for picking this up!  Getting a complete working armeb toolchain for real hardware can be involved[0,1].  That said, standard ARM GCC builds will generate object files with the `-mbig-endian` switch [2,3] even if the toolchain lacks the multilibs to generate complete programs.  You should be able to verify that this technique is sound with whatever GCC ARM is already in your CI system.

[0] https://bugs.launchpad.net/gcc-arm-embedded/+bug/1718206
[1] https://bugs.launchpad.net/gcc-arm-embedded/+bug/1469289
[2] https://www.godbolt.org/z/TWT4da6a6
[3] https://gcc.gnu.org/onlinedocs/gcc-10.3.0/gcc/ARM-Options.html

PR #195, opened 2021-09-01
Check output length argument size

DavidBurkett:
Currently, we check `out_len` against SIZE_MAX, but we haven't updated out_len yet. Instead, we should check the user provided argument `out_len_ll`


oconnor663:
Thank you! Apologies that I haven't maintained this file very well. It's only intended to be used for internal testing, so realistically these error paths probably aren't going to be triggered.

PR #201, opened 2021-10-07
Improve compile-time target detection for NEON

rsdy:
This PR moves most of the feature management for `neon` to the `build.rs` script. If a dependent crate enables the `neon` feature, that should gracefully compile also on all platforms even though they don't support NEON.

Additionally, I assume all `aarch64` targets to support NEON, and the optimization will turn on even if the feature isn't explicitly enabled.

I have a number of questions before this PR is ready for merge:

 1. Is enabling aarch64/NEON by default acceptable?
 2. Should we introduce a `no_neon` feature similar to how SSE/AVX is handled if the `aarch64` defaults to on?
 3. What's the best way to migrate C bindings and CI to cause the least amount of surprises?

### DIFF
--- a/c/blake3_impl.h
+++ b/c/blake3_impl.h
@@ -49,13 +49,18 @@ enum blake3_flags {
 #include <immintrin.h>
 #endif
 
-#if defined(IS_AARCH64) && !defined(BLAKE3_NO_NEON) && !defined(BLAKE3_USE_NEON)
-#define BLAKE3_USE_NEON
+#if !defined(BLAKE3_USE_NEON) 
+  // If BLAKE3_USE_NEON not manually set, autodetect based on AArch64ness
+  #if defined(IS_AARCH64)
+    #define BLAKE3_USE_NEON 1
+  #else
+    #define BLAKE3_USE_NEON 0
+  #endif
 #endif
 
 #if defined(IS_X86)
 #define MAX_SIMD_DEGREE 16
-#elif defined(BLAKE3_USE_NEON)
+#elif BLAKE3_USE_NEON == 1
 #define MAX_SIMD_DEGREE 4
 #else
 #define MAX_SIMD_DEGREE 1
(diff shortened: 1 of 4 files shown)


oconnor663:
This looks great. Thanks for making it so consistent with the x86 code. I'll want to play with it a little bit myself, but then I'll probably land it as-is.

> Is enabling aarch64/NEON by default acceptable?

I think we should. It's more in line with what we're already doing on x86, and it'll be a substantial speedup by default for a lot of users. (There was another PR about this at some point, but I think it got stalled in the review/feedback cycle.)

> Should we introduce a no_neon feature similar to how SSE/AVX is handled if the aarch64 defaults to on?

That sounds like a good idea, and we could add it to `.github/workflows/ci.yml` somewhere. Feel free to add that if you like, or we could also land this first and do that as a follow-up.

> What's the best way to migrate C bindings and CI to cause the least amount of surprises?

Do you mean the separate C implementation? I think it might be a good idea to make a similar change there. Maybe we could introduce a `-DBLAKE3_NO_NEON` flag, in line with `-DDBLAKE3_NO_SSE2` and others on x86.

I'm not _too_ worried about build compatibility here, since C callers probably vendor their own copy of our code, update that copy rarely (alas), and are more likely to read the patch notes when they do update. I think the impact here will be less than when we added SSE2 support. Also for what it's worth, we do say "Note that these steps may change in future versions" in the build docs, not that anybody reads the docs :)

c/blake3_impl.h
```diff
@@ -45,6 +45,10 @@ enum blake3_flags {
 #include <immintrin.h>
 #endif
 
+#if defined(__aarch64__) && !defined(BLAKE3_NO_NEON) && !defined(BLAKE3_USE_NEON)
```
oconnor663, line 48: Do we need to test `_M_ARM64` here in addition to `__aarch64__`, to cover MSVC?

oconnor663: Or maybe it would make sense to factor that into a separate variable, like we've done with `#if defined(__x86_64__) || defined(_M_X64) ` above?

oconnor663:
> The downside is the inconsistency with the x86 parts. But the inconsistency goes deeper than the macros, since there is no dynamic dispatch in place..

Yeah this is a good point. Might be worth calling this out explicitly in the build docs. I can add something.

@rsdy thanks for the revisions. I'm going to play with it on my RPi again, make sure CI is green, and then land it. Do you need the Rust code to be in a published release sooner rather than later? Or no rush?

oconnor663:
Landed! I'll make a note to cut a release in a week, on the off chance that anyone's actually testing master in the meantime :)

oconnor663:
Released as 1.1.0.

PR #209, opened 2021-11-10
Add a nix setup using flake

Anderssorby:
This adds a basic nix build which makes it possible to easily include BLAKE3 as a flake dependency.


oconnor663:
This feels like a lot of toplevel stuff to have in the repo, for what would be a small minority of users. Would it be possible to maintain it in a separate repo, or at least in a subdirectory?

Anderssorby:
@oconnor663 I can move it to my own repo. Though you should check out nix as a universal build tool.

Anderssorby:
I converted this to an independent solution.

Issue #212, opened 2021-11-23
digest v0.10


newpavlov:
In RustCrypto we plan to soon migrate our hash crates to digest v0.10, see https://github.com/RustCrypto/traits/pull/819 and https://github.com/RustCrypto/hashes/pull/217.

Probably the most notable change is that `crypto-mac` got merged into `digest`, so it should simplify a bit implementation of traits in your case. But most mid-level traits (`Update`, `FixedOutput`, etc.) haven't changed much, so migration should be relatively simple.

But I wonder if you will be interested in using the new block-level traits (see the [`core_api` module](https://github.com/RustCrypto/traits/blob/digest/v0.10/digest/src/core_api.rs)). By implementing them buffering will be handled automatically by the [`CoreWrapper`](https://github.com/RustCrypto/traits/blob/digest/v0.10/digest/src/core_api/wrapper.rs) type. There are also wrappers which handle switch between runtime and compile-time selection of output size. Notably, the [`block-buffer`](https://github.com/RustCrypto/utils/blob/block-buffer/v0.10/block-buffer/src/lib.rs) which is used by the wrapper types is designed to completely eliminate unreachable panics in buffering code without `unsafe` code in hash implementation crates.

If you have any comments, questions or suggestions about the new API, I will be happy to hear them out.

oconnor663:
Updating to digest v0.10 should be pretty natural as soon as that stabilizes. We've make a clear "no compatibility guarantees" note in our docs around this.

[...]

PR #215, opened 2021-12-22
Update digest crate to 0.10 for traits-preview feature

NeoRaider:
Adjust to the following changes that happened in digest:

- The crypto-mac crate has been merged into digest (with "mac" feature
  enabled)
- Various traits have been split up
- The Digest and Mac traits now share their update/finalize/reset
  implementations
- The BlockInput trait was dropped without replacement apparently (as
  long as the low-level core API is not used)

Related issue: #212 (not marked as closed by this PR, as discussion regarding the use the low-level API is still ongoing)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ std = ["digest/std"]
 # who use it should expect breaking changes between patch versions of this
 # crate. (The "*-preview" feature name follows the conventions of the RustCrypto
 # "signature" crate.)
-traits-preview = ["digest", "crypto-mac"]
+traits-preview = ["digest"]
 
 # ---------- Features below this line are for internal testing only. ----------
 
@@ -85,8 +85,7 @@ arrayvec = { version = "0.7.0", default-features = false }
 constant_time_eq = "0.1.5"
 rayon = { version = "1.2.1", optional = true }
 cfg-if = "1.0.0"
-digest = { version = "0.9.0", optional = true }
-crypto-mac = { version = "0.11.0", optional = true }
+digest = { version = "0.10.1", features = [ "mac" ], optional = true }
 
 [dev-dependencies]
 hex = "0.4.2"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,13 +49,12 @@
 //! without NEON support.
 //!
 //! The `traits-preview` feature enables implementations of traits from the
-//! RustCrypto [`digest`] and [`crypto-mac`] crates, and re-exports those crates
-//! as `traits::digest` and `traits::crypto_mac`. However, the traits aren't
-//! stable, and they're expected to change in incompatible ways before those
-//! crates reach 1.0. For that reason, this crate makes no SemVer guarantees for
-//! this feature, and callers who use it should expect breaking changes between
-//! patch versions. (The "-preview" feature name follows the conventions of the
-//! RustCrypto [`signature`] crate.)
+//! RustCrypto [`digest`] crate, and re-exports those crates as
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -85,8 +85,7 @@ arrayvec = { version = "0.7.0", default-features = false }
 constant_time_eq = "0.1.5"
 rayon = { version = "1.2.1", optional = true }
 cfg-if = "1.0.0"
-digest = { version = "0.9.0", optional = true }
-crypto-mac = { version = "0.11.0", optional = true }
+digest = { version = "0.10.1", features = [ "mac" ], optional = true }
```
oconnor663, line 88: Not a problem with this PR, but I think we have an issue where `std = ["digest/std"]` on line 26 unfortunately enables the digest feature all the time. So `optional = true` here isn't really behaving like we'd expect it to behave. Do you know of any way to solve this? Maybe we should just get rid of `std = ["digest/std"]`?

src/lib.rs
```diff
@@ -49,13 +49,12 @@
...
-//! RustCrypto [`digest`] and [`crypto-mac`] crates, and re-exports those crates
-//! as `traits::digest` and `traits::crypto_mac`. However, the traits aren't
-//! stable, and they're expected to change in incompatible ways before those
-//! crates reach 1.0. For that reason, this crate makes no SemVer guarantees for
-//! this feature, and callers who use it should expect breaking changes between
-//! patch versions. (The "-preview" feature name follows the conventions of the
-//! RustCrypto [`signature`] crate.)
+//! RustCrypto [`digest`] crate, and re-exports those crates as
+//! `traits::digest`. However, the traits aren't stable, and they're expected to
+//! change in incompatible ways before those crates reach 1.0. For that reason,
```
oconnor663, line 54: `s/those crates/that crate/`

src/lib.rs
```diff
@@ -901,8 +899,7 @@ fn parent_node_output(
 ///
 /// When the `traits-preview` Cargo feature is enabled, this type implements
 /// several commonly used traits from the
-/// [`digest`](https://crates.io/crates/digest) and
-/// [`crypto_mac`](https://crates.io/crates/crypto-mac) crates. However, those
+/// [`digest`](https://crates.io/crates/digest) crate. However, those
 /// traits aren't stable, and they're expected to change in incompatible ways
 /// before those crates reach 1.0. For that reason, this crate makes no SemVer
```
oconnor663, line 904: same

oconnor663:
Thank you! This will save me a lot of time looking through the change notes :)

Cargo.toml
oconnor663: Oops never mind. I just played with it, and while it's true that the `digest` dependency does get pulled in by default (no biggie), our `traits-preview` feature is _not_ enabled by default. So callers still can't rely on these trait implementations without explicitly asking for the feature, which is what we really care about for (in)stability purposes.

newpavlov:
>The BlockInput trait was dropped without replacement apparently

It was replaced with the [`digest::crypto_common::BlockSizeUser`](https://docs.rs/crypto-common/0.1.1/crypto_common/trait.BlockSizeUser.html) trait. The [`hmac::SimpleHmac`](https://docs.rs/hmac/latest/hmac/struct.SimpleHmac.html) type uses it to get the block size used in HMAC. Note that you can no longer use [`hmac::Hmac`](https://docs.rs/hmac/latest/hmac/type.Hmac.html) with BLAKE2/BLAKE3 since it now works only with hashes which expose block-level interface and consume blocks eagerly.

PR #228, opened 2022-03-04
publish reference implementation as a crate

chippers:
Hi, I'm making this as a PR instead of an issue just so that I can kick-start this potentially happening, feel free to treat it more like an issue.

starting off with a few things related to the actual code changes:
1. I intentionally kept the file structure the same
2. I made the version as `1.0.0` because the main crate is already major version 1 and this reference seems rather stable change-wise
3. I licensed it same as the main crate explicitly since it was previously ambiguous
4. I added the clippy commit separately so that it can be reverted easily if wanted
    * `Default` for `Hasher`
    * `u64` enumeration with `.zip()` instead of manual counter
5. I used underscore instead of a hypen since it seems like other blake3 stuff leans towards that

# why a crate?

I've found that using the reference implementation is very desirable when using it inside build-time scripts and codegen dependencies. Even though the code is slower during runtime, it takes a magnitude longer to compile. This becomes especially exasperated when building alongside a lot of other dependencies and `blake3` becomes a single unit being compiled alongside many other crates.

Because the runtime during codegen crates is also compile time, sacrificing small runtime performance penalties while gaining a magnitude more time during the original compilation usually results in large compile time gains. Of course this depends on what the codegen is doing, but in our case we are bundling already-built web front-ends to embed in our resulting binary while using `blake3` to prevent extra preprocessing work during dirty builds. This typically means number of iterations is well within a couple dozen files.

I measured this on a CPU constrained computer to show the effect of this easier (ThinkPad T410 Intel i5-520M 2c/4t). The following runtime timings come from this laptop using `hyperfine` and the compile times from the `--timings` flag on cargo.

Runtime (a simplified custom b3sum utility ran on a 6.8MiB JavaScript file):
<details>

<summary>The custom b3sum code</summary>

Notably, I used the same code for both crates, with exception of not needing to create the `[u8; 32]` manually.

```rust
use crate::reference::Hasher;
use std::env::args_os;
use std::fmt::Write;
use std::fs::File;
use std::io::{self, Read};

mod reference;

fn copy_wide(mut reader: impl Read, hasher: &mut reference::Hasher) -> io::Result<u64> {
    let mut buffer = [0; 65536];
    let mut total = 0;
    loop {
        match reader.read(&mut buffer) {
            Ok(0) => return Ok(total),
            Ok(n) => {
                hasher.update(&buffer[..n]);
                total += n as u64;
            }
            Err(ref e) if e.kind() == io::ErrorKind::Interrupted => continue,
            Err(e) => return Err(e),
        }
    }
}

fn main() {
    let input = args_os().nth(1).expect("at least 1 argument of file path");
    let file = File::open(&input).expect("unable to open up input file path");
    let _copied = copy_wide(file, &mut hasher).expect("io error while copying file to hasher");
    let mut output = [0u8; 32];
    hasher.finalize(&mut output);
    let mut s = String::with_capacity(2 * output.len());
    for byte in output {
        write!(s, "{:02x}", byte).expect("cant write hex byte to hex buffer");
    }
    println!("{} {}", s, input.to_string_lossy());
}
```

</details>

| crate | features | compile time | run time |
| --- | --- | --- | --- |
| `blake3` | `rayon` | 19.2s | 8.7ms |
| `blake3` | | 8.5s | 14.9ms |
| `blake3_reference` | | 1.5s | 37.9ms |

This is the best-case scenario, compilation times where only that small custom b3sum is using CPU. Even so, it would take ~600 items for the rayon version to catch up with the reference version in a codegen situation and around ~300 for the non-rayon version.

Unfortunately, this doesn't behave linearly when thrown in with a lot of other dependencies compiling at the same time. Due to out other big dependency also building C code during its build script (`zstd`), these happen to be exasperated. The following is the timings (compile time only) while building [Tauri's API example](https://github.com/tauri-apps/tauri/tree/dev/examples/api).
| crate | profile | compile time |
| --- | --- | --- |
| `zstd-sys build script (run)` | `dev` | 215.88s |
| `blake3 build script (run)` | `dev` | 143.89s |
| --- | --- | --- |
| `zstd-sys build script (run)` | `release` | 243.98s |
| `blake3 build script (run)` | `release` | 145.75s |

I compare against `zstd` (who is impacted even worse by this) because these 2 build scripts take up a significant chunk of the compile time. We are also looking to replace `zstd` with something that isn't compiling C to alleviate this. These large compile times is really what caused me to look into this and realize that the reference is a great balance for codegen crates depending on their characteristics.

Notably, the reference seems to not have a problem and still compiles in 1.5s in both profiles (due to the simple code). So if we were to do our earlier comparison of how many hashes would need to be generated to be worth it we would arrive at (using the dev profile) at just under ~4900 items for the rayon version compared to the reference.

I would also like to point out that the item in this case is `6.8MiB`, and really represents X amount of times of bytes rather than files. Since this file was the vendor file of a large JavaScript application, it is also the majority of the size for most applications so it could also be seen that we are saying X times in size of the application rather than individual files.

## again, why the crate?

This doesn't really **need** to be a crate, but I figured this was better than manually vendoring. We would get any reference changes (checking the git history they are typically small and long in-between with little logic changes) in the future without needing to worry about being potentially out-of-date. I also believe that this could be useful for other projects in our position where they use `blake3` during codegen, and so runtime and compile time become additive.

---

Let me know if there are any changes you want me to make or any thoughts/questions you might have. Feel free to change anything in this PR before merging.


reference_impl/reference_impl.rs
```diff
@@ -382,3 +380,9 @@ impl Hasher {
         output.root_output_bytes(out_slice);
     }
 }
+
+impl Default for Hasher {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
oconnor663, line 388: One of the reasons I've leaned against publishing this implementation on crates.io is that there's a tension between keeping it simple and making it actually usable. In addition to `Default`, a real `Hasher` should also implement `Debug` and `Clone` and probably `Write`. But on the other hand, the primary goal of `reference_impl.rs` is to be readable, including to people who don't actually know Rust. I think adding common traits into the mix detracts from that, because less experienced readers don't know what `#[derive(...)]` means, and it's not clear that they can gloss over it.

oconnor663:
I'd like to understand these build time better. What is it that's taking 100+ seconds? It might be that you're thrashing your laptop's disk, or exhausting RAM and leaning on swap, or something like that. Is there any chance `cargo build --jobs=1` does better in that environment?

reference_impl/reference_impl.rs
```diff
@@ -154,8 +154,7 @@ impl Output {
     }
 
     fn root_output_bytes(&self, out_slice: &mut [u8]) {
-        let mut output_block_counter = 0;
-        for out_block in out_slice.chunks_mut(2 * OUT_LEN) {
+        for (output_block_counter, out_block) in (0u64..).zip(out_slice.chunks_mut(2 * OUT_LEN)) {
```
oconnor663, line 157: This is another place where I want to optimize for people being able to read this without necessarily knowing Rust. I do admit that I use `zip` in `words_from_little_endian_bytes` above, but that helper isn't really specific to BLAKE3, and hopefully the name makes it obvious what it does.

chippers:
i agree with both those comments, which is why i kept the clippy stuff separate. i've reverted them.

ill try the timings again with limited jobs.

For this test it's a T410 laptop with 2c/4t, 8GiB RAM, and 120GiB SSD. We also have a windows example with a 4c CPU at tauri-apps/tauri#3571

chippers:
These timings are a bit surprising to me and are revealing perhaps a bit more on how cargo handles jobs for build scripts. Running `--jobs=1` made the dev build time from ~8m35s to ~18m, which is relatively expected. Unexpectedly, the length of build scripts also rose rather significantly. According to the [timings chart](https://gist.github.com/chippers/369f24ab99ecf8af9542e7bb5d9d3cfc), these all rose in parallel

| build script | old | `jobs=1` |
| --- | --- | --- |
| `ring` | 91s | 442.9s |
| `zstd-sys` | 215.88s | 526.57s |
| `blake3` | 143.89s | 451.19s |

This is making me even more suspicious that a large part of this exaggeration of compile times compared to simple projects is that multiple C-compiling build scripts eat each other heavily. I'm going to write up a quick branch that changes `blake3` to the reference and removes ring (since this example doesn't use it). Ideally, we see the times for `zstd-sys` drop to normal levels since it's the only one competing for CPU.

Additionally, maybe `blake3_reference` isn't the answer that's right for this project. I was leaning towards in-tree support for it, but with the infrequency of changes to the reference, I think another good solution is publishing my own `blake3-simple` crate that lightly modifies the interfaces of the reference while keeping the core the same.

chippers:
Ok, I made a branch that did not use `ring` or `blake3`. I did **not** use `--jobs=1`.
| build script | old | new |
| --- | --- | --- |
| `zstd-sys` | 215.88s | 149.89 |

So it seems likely that having multiple dependencies that build C code is the true culprit, and not `blake3` by itself (which was my thought from the start). Since `blake3` _does_ build some C code, I figure that having the reference as a crate would be an easy solution to that.

Let me know if you think the `blake3-simple` idea from my previous comment is more appropriate, I just leaned towards upstream first.

PR #240, opened 2022-04-09
add prototypes to fix -Wstrict-prototypes warnings

MaskRay:
---
BLAKE3 has been imported into llvm-project. This patch fixes warnings when llvm-project is compiled by recent Clang which enables `-Wstrict-prototypes` under (the default) `-pedantic`. In case of an upgrade, llvm-project can avoid the local patch: https://github.com/llvm/llvm-project/commit/7b0dad9a0204604bd72a5af4c94ba31c5133f71f

### DIFF
--- a/c/README.md
+++ b/c/README.md
@@ -13,7 +13,7 @@ result:
 #include <string.h>
 #include <unistd.h>
 
-int main() {
+int main(void) {
   // Initialize the hasher.
   blake3_hasher hasher;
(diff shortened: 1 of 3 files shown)


oconnor663:
Interesting, today I learned: https://stackoverflow.com/questions/693788/is-it-better-to-use-c-void-arguments-void-foovoid-or-not-void-foo

PR #243, opened 2022-04-17
Use weak dependency feature for 'digest/std'

jsimmons:
This is stable now as of Rust 1.60, not sure whether you want to have that as min-spec, though.

Fixes BLAKE3-team/BLAKE3/204


elichai:
@oconnor663 This can be closed now that https://github.com/BLAKE3-team/BLAKE3/pull/309 removed the `digest/std` dependency entirely

PR #251, opened 2022-07-18
Add link to README MT section

Rudxain:
This makes it easier to go to the explanation of why it isn't multithreaded


oconnor663:
Thanks!

oconnor663:
When I test this, it seems like the anchor works if I open the link in a new tab, but not if I click through from the project root. Please let me know if you see a way to fix this.

Rudxain:
> It seems like the anchor works if I open the link in a new tab, but not if I click through from the project root

Same happens to me, both on mobile-web and desktop-web. It seems like a recent GH bug (it affects other anchors in other repos too), because IIRC it wasn't like that some months ago

PR #255, opened 2022-09-09
Enable optimisation in Cargo.toml of b3sum

Yusuto:
Enable [stripping](https://doc.rust-lang.org/cargo/reference/profiles.html#strip), [link-time optimization](https://doc.rust-lang.org/cargo/reference/profiles.html#lto) and and all compiler [optimizations](https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level). This should make executables considerably smaller (~800kB instead of >4MB for Linux, through stripping and lto, could be improved by setting `opt-level = "s"`, see [The Cargo Book](https://doc.rust-lang.org/cargo/reference/profiles.html?highlight=opt-level#opt-level)) and maybe faster (through lto and optimizations, although my dodgy tests didn't show any speed increase, could be because of a hard-drive bottleneck which would honestly be insane how fast this algorithm is).

I dun know yet how to exactly test performance of such performance-critical programs, the best I could to is make a sloppy test by writing like 10GiB of `/dev/urandom` to a file and then `time`ing the execution of `b3sum`, but I'm honestly super not trusting this method:
```bash
$ du test
10929668
$ time ./target/release/b3sum test
6edc4eb802feb666e4ee9fc388eab08c39906bc5d455ed14739ec6742017a47a  test
./target/release/b3sum test  2.18s user 4.83s system 187% cpu 3.735 total
$ c clean
$ c b --release
$ time ./target/release/b3sum test
6edc4eb802feb666e4ee9fc388eab08c39906bc5d455ed14739ec6742017a47a  test
./target/release/b3sum test  2.10s user 4.76s system 186% cpu 3.679 total
``` 

I also checked compile times, and they are staying very reasonable:
```
Before:
cargo b --release  51.84s user 3.73s system 562% cpu 9.883 total
After:
cargo b --release  36.13s user 3.18s system 300% cpu 13.088 total
```
Wait, why is it so much faster? I am- mildly confused...

Is there a reason this stuff hasn't been added yet?

### DIFF
--- a/.github/workflows/build_b3sum.py
+++ b/.github/workflows/build_b3sum.py
@@ -9,7 +9,7 @@ import sys
 ROOT = Path(__file__).parent.parent.parent
 RUST_TARGET = sys.argv[1]
 
-subprocess.run(["cargo", "build", "--target", sys.argv[1], "--release"],
+subprocess.run(["cargo", "build", "--target", sys.argv[1], "--release", "--config", "strip=true"],
                cwd=ROOT / "b3sum")
 
 if platform.system() == "Windows":
--- a/b3sum/Cargo.toml
+++ b/b3sum/Cargo.toml
@@ -25,3 +25,8 @@ wild = "2.0.3"
 [dev-dependencies]
 duct = "0.13.3"
 tempfile = "3.1.0"
+
+[profile.release]
+strip = true
+lto = true
+opt-level = 3


oconnor663:
Can we compare this to what some popular Rust binary crates are doing? In particular, it looks like `ripgrep` [doesn't turn on these options](https://github.com/BurntSushi/ripgrep/blob/a66315d232b49e623268de801af19460108ba514/Cargo.toml#L79-L80). Maybe the idea is that individual packagers can turn on the options they prefer using environment variables? @burntsushi I'd be curious to get your advice about this. I don't see much when I search for "rust lto best practices" or similar.

oconnor663:
Regarding performance testing, I don't expect any of these flags to have much impact in the default case either way. The most performance-critical parts are written in assembly, which won't be affected by these flags. However, it might be interesting to test the pure/portable Rust implementation, by measuring with `--features pure`.

Yusuto:
From what I could see, compile times are largely unaffected here by LTO, but as you said so is performance. I think the things that I found more peculiar were stripping and optimization, especially the former, since it makes the file four times smaller, which I think is nice. For the other two I think I agree with »If it doesn't matter, why bother?«.

Yusuto:
Should we maybe just remove LTO and Compiler Optimizations and only enable stripping? That part just seems like a free 75% reduction in file size, right?

oconnor663:
How about we enable stripping in our internal CI build (the binaries that get hosted here on GitHub and attached to each release tag) but keep the defaults in `Cargo.toml`?

Yusuto:
I think that actually sounds like a very good idea, Yes!
That can probably be easily automated, right? There is probably some option for enabling stripping on the command-line, something like `cargo build --release --profile strip=true` (that doesn't actually work, sadly) in the CI?

oconnor663:
Looks like we might want to copy something like what `ripgrep` does here: https://github.com/BurntSushi/ripgrep/blob/4386b8e805e273a9795ad4e256c455c74407c949/.github/workflows/release.yml#L135-L146

And put it in between these two steps here: https://github.com/BLAKE3-team/BLAKE3/blob/7e15c5314ea0ca34c52fd39583b1aa7c07eac1ae/.github/workflows/tag.yml#L38-L45

Yusuto:
Hmm, couldn't we just let Cargo take care of stripping, now that the feature is stabilized? I finally got to look up how, it seems to be `cargo build --release --config strip=true`, and that would need to be changed in [`build_b3sum.py`](https://github.com/BLAKE3-team/BLAKE3/blob/7e15c5314ea0ca34c52fd39583b1aa7c07eac1ae/.github/workflows/build_b3sum.py#L12)

Issue #220, opened 2022-01-13
data race in get_cpu_features()


arvidn:
according to https://github.com/BLAKE3-team/BLAKE3/issues/99 blake3 is supposed to be safe to use in multi-threaded environments.

However, the [get_cpu_features](https://github.com/BLAKE3-team/BLAKE3/blob/master/c/blake3_dispatch.c#L85) implementation reads from and writes to the global variable `g_cpu_features` *without* synchronization. This triggers an error in thread sanitizer.

This function is called from `blake3_compress_in_place()`, `blake3_compress_xof()`, `blake3_hash_many()` and `blake3_simd_degree()`.

There are a few ways to solve this:

1. document that these functions are not safe to call from multiple different threads
2. make `g_cpu_features` thread local (with the `_Thread_local` storage class specifier)
3. use `call_once()` to ensure `g_cpu_features` is initialized exactly once, and synchronize with all other threads

It seems (3) is most consistent with the intention of the code. Here's a proposed patch to do this:

```
--- a/c/blake3_dispatch.c
+++ b/c/blake3_dispatch.c
@@ -1,6 +1,7 @@
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
+#include <threads.h>
 
 #include "blake3_impl.h"
 
@@ -71,20 +72,25 @@ enum cpu_feature {
   UNDEFINED = 1 << 30
 };
 
+static once_flag g_cpu_features_initialized = ONCE_FLAG_INIT;
+
 #if !defined(BLAKE3_TESTING)
 static /* Allow the variable to be controlled manually for testing */
 #endif
     _Atomic enum cpu_feature g_cpu_features = UNDEFINED;
 
+static void set_cpu_features_impl();
+
 #if !defined(BLAKE3_TESTING)
 static
 #endif
     enum cpu_feature
     get_cpu_features() {
+  call_once(&g_cpu_features_initialized, set_cpu_features_impl);
+  return g_cpu_features;
+}
 
-  if (g_cpu_features != UNDEFINED) {
-    return g_cpu_features;
-  } else {
+static void set_cpu_features_impl() {
 #if defined(IS_X86)
     uint32_t regs[4] = {0};
     uint32_t *eax = &regs[0], *ebx = &regs[1], *ecx = &regs[2], *edx = &regs[3];
@@ -123,13 +129,11 @@ static
       }
     }
     g_cpu_features = features;
-    return features;
 #else
[...]

PR #256, opened 2022-09-10
blake3_dispatch: Fix race condition initializing g_cpu_features.

jblazquez:
If multiple threads try to compute a hash simultaneously before the library has been used for the first time, the logic in `get_cpu_features` that detects CPU features will write to `g_cpu_features` without synchronization, which is a race condition, thus undefined behavior (and flagged by ThreadSanitizer).

This change marks `g_cpu_features` as an atomic variable to address the race condition. Since the C library is compiled with `-std=c11` this doesn't seem to introduce any compatibility issues.

I understand that the C library is not currently multithreaded (as documented [here](https://github.com/BLAKE3-team/BLAKE3/blob/master/c/README.md#multithreading)), but I understand that to mean that it doesn't support speeding up the hashing of a single buffer by using multiple threads, not that the library can't be used in a multithreaded application at all. Right now, despite this race condition being "benign" (all threads compute the same value, so the setting of `g_cpu_features` is idempotent), my understanding is that it does represent undefined behavior, so it's something that should be avoided. And it triggers ThreadSanitizer warnings, which can be suppressed but ideally would not occur at all.

Repro steps:

1. Create a file called `blake_tsan_error.c` at the root of the repo:

```c
#include <blake3.h>
#include <pthread.h>

void* thread_func(void* p) {
    uint8_t hash[BLAKE3_OUT_LEN];
    blake3_hasher hasher;
    blake3_hasher_init(&hasher);
    blake3_hasher_finalize(&hasher, hash, sizeof(hash));
    return NULL;
}

int main(int argc, char** argv) {
    pthread_t thread1;
    pthread_create(&thread1, NULL, &thread_func, NULL);
    pthread_t thread2;
    pthread_create(&thread2, NULL, &thread_func, NULL);
    pthread_join(thread1, NULL);
    pthread_join(thread2, NULL);
}
```

2. Build the code using Clang 15:

```shell
export CFLAGS="-O3 -std=c11 -pedantic -fstack-protector-strong -D_FORTIFY_SOURCE=2 -fPIE -fvisibility=hidden -fsanitize=thread -fno-omit-frame-pointer -fno-optimize-sibling-calls -DBLAKE3_NO_SSE2 -DBLAKE3_NO_SSE41 -DBLAKE3_NO_AVX2 -DBLAKE3_NO_AVX512"

clang $CFLAGS -c blake3.c -o blake3.o
clang $CFLAGS -c blake3_dispatch.c -o blake3_dispatch.o
clang $CFLAGS -c blake3_portable.c -o blake3_portable.o
clang $CFLAGS -I$PWD/c blake_tsan_error.c *.o -o blake_tsan_error
```

3. Run the code:

```shell
$ TSAN_OPTIONS=halt_on_error=1 ./blake_tsan_error
==================
WARNING: ThreadSanitizer: data race (pid=47788)
  Read of size 4 at 0x55981bb27c18 by thread T2:
    #0 blake3_compress_xof <null> (blake_tsan_error+0xece78)
    #1 blake3_hasher_finalize_seek <null> (blake_tsan_error+0xeb982)
    #2 blake3_hasher_finalize <null> (blake_tsan_error+0xeb6bb)
    #3 thread_func <null> (blake_tsan_error+0xe8481)

  Previous write of size 4 at 0x55981bb27c18 by thread T1:
    #0 blake3_compress_xof <null> (blake_tsan_error+0xecf16)
    #1 blake3_hasher_finalize_seek <null> (blake_tsan_error+0xeb982)
    #2 blake3_hasher_finalize <null> (blake_tsan_error+0xeb6bb)
    #3 thread_func <null> (blake_tsan_error+0xe8481)

  Location is global 'g_cpu_features' of size 4 at 0x55981bb27c18 (blake_tsan_error+0x12cc18)

  Thread T2 (tid=47791, running) created by main thread at:
    #0 pthread_create tsan_interceptors_posix.cpp:1022:3 (blake_tsan_error+0x5f933)
    #1 main <null> (blake_tsan_error+0xe84ff)

  Thread T1 (tid=47790, finished) created by main thread at:
    #0 pthread_create tsan_interceptors_posix.cpp:1022:3 (blake_tsan_error+0x5f933)
    #1 main <null> (blake_tsan_error+0xe84ec)

SUMMARY: ThreadSanitizer: data race (blake_tsan_error+0xece78) in blake3_compress_xof
==================
```

### DIFF
--- a/c/blake3_dispatch.c
+++ b/c/blake3_dispatch.c
@@ -6,12 +6,39 @@
 
 #if defined(IS_X86)
 #if defined(_MSC_VER)
+#include <Windows.h>
 #include <intrin.h>
 #elif defined(__GNUC__)
 #include <immintrin.h>
 #else
 #error "Unimplemented!"
 #endif
+#endif
+
+#if !defined(BLAKE3_ATOMICS)
+#if defined(__has_include)
+#if __has_include(<stdatomic.h>) && !defined(_MSC_VER)
+#define BLAKE3_ATOMICS 1
+#else
+#define BLAKE3_ATOMICS 0
+#endif /* __has_include(<stdatomic.h>) && !defined(_MSC_VER) */
+#else
+#define BLAKE3_ATOMICS 0
+#endif /* defined(__has_include) */
+#endif /* BLAKE3_ATOMICS */
+
+#if BLAKE3_ATOMICS
+#define ATOMIC_INT _Atomic int
+#define ATOMIC_LOAD(x) x
+#define ATOMIC_STORE(x, y) x = y
+#elif defined(_MSC_VER)
+#define ATOMIC_INT LONG
+#define ATOMIC_LOAD(x) InterlockedOr(&x, 0)
+#define ATOMIC_STORE(x, y) InterlockedExchange(&x, y)
+#else
+#define ATOMIC_INT int
+#define ATOMIC_LOAD(x) x
+#define ATOMIC_STORE(x, y) x = y
 #endif
 
 #define MAYBE_UNUSED(x) (void)((x))
@@ -76,7 +103,7 @@ enum cpu_feature {
 #if !defined(BLAKE3_TESTING)
 static /* Allow the variable to be controlled manually for testing */
 #endif
-    enum cpu_feature g_cpu_features = UNDEFINED;
+    ATOMIC_INT g_cpu_features = UNDEFINED;
 
 #if !defined(BLAKE3_TESTING)
 static
@@ -84,14 +111,16 @@ static
     enum cpu_feature
     get_cpu_features(void) {
 
-  if (g_cpu_features != UNDEFINED) {
-    return g_cpu_features;
+  /* If TSAN detects a data race here, try compiling with -DBLAKE3_ATOMICS=1 */
+  enum cpu_feature features = ATOMIC_LOAD(g_cpu_features);
+  if (features != UNDEFINED) {
+    return features;
   } else {
 #if defined(IS_X86)
     uint32_t regs[4] = {0};
     uint32_t *eax = &regs[0], *ebx = &regs[1], *ecx = &regs[2], *edx = &regs[3];
     (void)edx;
-    enum cpu_feature features = 0;
+    features = 0;
     cpuid(regs, 0);
     const int max_id = *eax;
     cpuid(regs, 1);
@@ -124,7 +153,7 @@ static
         }
       }
     }
-    g_cpu_features = features;
+    ATOMIC_STORE(g_cpu_features, features);
     return features;
 #else
     /* How to detect NEON? */


c/blake3_dispatch.c
```diff
@@ -76,22 +77,32 @@ enum cpu_feature {
 #if !defined(BLAKE3_TESTING)
 static /* Allow the variable to be controlled manually for testing */
 #endif
-    enum cpu_feature g_cpu_features = UNDEFINED;
+#if defined(_MSC_VER)
+LONG
+#else
+_Atomic int
```
jblazquez, line 83: MSVC doesn't support `_Atomic` yet when building C code (https://github.com/microsoft/STL/blob/main/stl/inc/stdatomic.h#L15), so the Win32 interlocked APIs have to be used.

oconnor663:
We considered something similar in https://github.com/BLAKE3-team/BLAKE3/issues/220 but decided against it for compatibility reasons. Standard atomics don't seem to be widely supported enough yet, and there are lots of callers who want to build this code on super old platforms. The fact that this undefined behavior is confined to x86 also makes me feel slightly less bad about it, even though I agree it's bad.

jblazquez:
I see. Thanks for pointing me to the existing issue.

We’re OK keeping this as a local patch, but hopefully the library can address this in the future.

oconnor663:
Maybe a good middle ground could be to hide standard atomics behind a platform independent preprocessor flag like `BLAKE3_ATOMICS` or something. Users who turn on that flag get the atomic version (let's say with `memory_order_relaxed` for the loads), but by default we use the current version. We could put comments on the relevant lines saying something like "Hi, did `-fsanitize=thread` bring you here? If so you can set `-DBLAKE3_ATOMICS=1` to make this error go away. But be aware that not all compilers support standard atomics, particularly not MSVC as of 2022."

How does that sound? I'm eager to avoid platform-specific reasoning about who supports what, but at the same time I do want to support `-fsanitize=thread`.

jblazquez:
That’s not a bad idea. And actually, we may not even need to do that, since we could use the compiler’s __has_include functionality to ask whether the <stdatomic.h> header exists, and if so that’s a good indication that C11 atomics are supported.

Except for MSVC, which would still require the use of the InterlockedXXX functions for a while, like my PR does. The good thing is that MSVC and Windows have supported these forever.

If this sounds good I can update my PR and test against some older compilers in a Docker container.

elichai:
FWIW in x86 you can just use an inline assembly with a `mov` instruction.
Another option is using `_Threadlocal` but that's probably has the same support as `_Atomic`

jblazquez:
@oconnor663, what do you think of the latest changes? I tested building with [GCC 4.6](https://gcc.gnu.org/wiki/C11Status) (the oldest compiler that supports C11), GCC 4.9 (the first one to support `_Atomic`), Clang 3.0, and MSVC 2022.

sneves:
[This](https://c.godbolt.org/z/c4GEGKKEW) works almost everywhere, from clang >= 3.1, gcc >= 4.1.2, MSVC >= 2015 (I think >= 2010 will work, but cannot verify), ICC >= 13.0. It falls back to the 'unsafe' code when it's not one of these. Really, this is just an elaborate rigmarole to generate the exact same code any compiler will generate anyway (on x86).

`_Atomic` seems more trouble than it's worth; on compilers like ICC 17, the compiler will tell you it supports C11, has `stdatomic.h`, but then will not recognize what `_Atomic` is.

sneves:
Codegen indicates yes; `__tsan_atomic32_{load,store}` are emitted instead of `__tsan_{read,write}4` wherever `-fsanitize=thread` is supported. I haven't run it.

jblazquez:
@oconnor663, coming back to this PR since I realized had not been merged and I was hoping to upgrade to 1.4.1

Do you think the PR is in a state where it can be merged?

Thanks.

oconnor663:
@jblazquez thanks for all your work on this, and apologies for taking so long to review it. This looks good to me, and I'm going to land it as-is. If we want to modify it from here to work around ICC or whatever other lying-about-supporting-standard-headers compiler, this PR makes it pretty easy to add additional cases to the `!defined(_MSC_VER)` clause, and that feels good enough.

I'm about to go off the grid for a month, so if this causes unexpected issues, @sneves please feel free to revert without waiting for me :)

PR #258, opened 2022-09-14
Add vcpkg installation instructions

Cheney-W:
`blake3` available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for `blake3` and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build `blake3`, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/blake3/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,18 @@ time openssl sha256 /tmp/bigfile
 time b3sum /tmp/bigfile
 ```
 
+### Installing from vcpkg
+
+You can download and install `blake3` using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+```sh
+git clone https://github.com/Microsoft/vcpkg.git
+cd vcpkg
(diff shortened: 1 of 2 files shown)


oconnor663:
It might make more sense to include something like this in `c/README.md` than in the toplevel readme.

By the way, your CMakeLists.txt looks great. Is it portable? Could it make sense to upstream it into this repo?

Cheney-W:
Thanks for your reply! I will move this instruction into the `c/README.md`.
About the CMakeLists.txt, I don't quite understand what exactly you mean by `portable`, but for our use, it works fine on windows, linux and osx, so I it make sense to upstream it into this repo.

PR #259, opened 2022-09-14
Rewrite CLI using Derive API

sorairolake:
This makes the code easier to read, write and modify.

ref: [When should I use the builder vs derive APIs?](https://docs.rs/clap/latest/clap/_faq/index.html#when-should-i-use-the-builder-vs-derive-apis)

### DIFF
--- a/b3sum/Cargo.lock
+++ b/b3sum/Cargo.lock
@@ -105,6 +105,7 @@ dependencies = [
  "once_cell",
  "strsim",
  "termcolor",
+ "terminal_size",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


oconnor663:
Thank you, I've been meaning to do this for a while. Now that v4 is out, do you think it makes sense to go ahead and upgrade at the same time?

sorairolake:
Bump clap to v4.0.
Also, change to print the summary only, when short help (`-h`) is specified.

oconnor663:
Sorry for the delay in merging this, and thanks so much for saving me from reading through the clap v4 changelog 😅

PR #263, opened 2022-10-04
Add #ifdef guards to enable multi-arch compilation on macOS

patstew:


oconnor663:
cc @sneves

oconnor663:
I've had to go through [a lot of trouble](https://github.com/oconnor663/blake3-py/blob/master/c_impl/setup.py#L98-L166) to get another project to build this code into a universal binary for macOS, and if this is the better way to get that working, I'm all for it.

PR #267, opened 2022-11-12
constant time Ord and PartialOrd implementations

### LINKED ISSUE #202: Hash struct lacks an Ord implementation
Without an `Ord` implementation (or even a `Hash` implementation) `blake3::Hash` values cannot be used as keys in `BTreeMap`s or `HashMap`s, among many other valid uses for (vartime) orderings.

To be safe; these implementations ought to be constant-time. Either way, the [subtle](https://docs.rs/subtle) crate provides traits which are explicitly intended for constant time equality and ordering.

As a workaround, I'm using the newtype pattern to implement these traits on my own `OrdHash` type.

If a PR would be welcome for this addition, should the implementation be constant time or variable?

imuli:
Fixes #202.

This does not follow the recommended path of utilizing [subtle](https://docs.rs/subtle/latest/subtle/trait.ConstantTimeLess.html) as it does not provide any helpers to do constant time ordering on containers.

The overall method is to compare each u32 in the hashes, turn each of those into -1, 0, or 1, accumulate them into a single signed word, and then finally compare that word to 0.

The [optimized object code](https://github.com/BLAKE3-team/BLAKE3/files/9993917/disassembly.txt) on x86-64 is somewhat large but appears to be constant time - no jumps, early returns, or anything like that. This is, of course, not a guarantee that it will never face optimization in hardware or software in the future.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -304,6 +304,46 @@ impl PartialEq<[u8]> for Hash {
 
 impl Eq for Hash {}
 
+/// This implementation is constant-time.
+impl PartialOrd for Hash {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+/// This implementation is constant-time.
+impl Ord for Hash {
+    #[inline]
+    fn cmp(&self, other: &Hash) -> std::cmp::Ordering {
+        let self32: [u32; 8] = as_u32s(self);
+        let other32: [u32; 8] = as_u32s(other);
+        let mut acc: i32 = 0;
+        for i in 0..self32.len() {
+            // the left shift keeps earlier comparisons more significant than later ones
+            acc = (acc<<1) + cmp_sign(&self32[i], &other32[i]);
+        }
+        acc.cmp(&0)
+    }
+}
+
+/// Interpret a Hash as an array of big endian 32 bit unsigned integers.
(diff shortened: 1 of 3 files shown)


imuli:
> * Like we discussed in [Hash struct lacks an Ord implementation #202](https://github.com/BLAKE3-team/BLAKE3/issues/202), it's not clear that we _want_ a constant time `Ord` implementation, because many (most?) practical use cases for `Ord` will silently break that constant-time guarantee. For experienced callers who know exactly what they're doing, the conversion to `&[u8; 32]` is no problem.
Yeah, given that discussion I'd be inclined toward the original PR.

> (Side note: the CI failures are because we test this crate in `no_std` mode, where `std::cmp` doesn't exist. We already use `core::cmp` at the top of `lib.rs`, so the fix would be to just remove the `std::` prefix. You can test this yourself with something like `cargo check --no-default-features`.)
That makes a lot of sense!

oconnor663:
> it's not clear that we _want_ a constant time `Ord` implementation

Actually I should've said that it's not clear (to me) that we want any `Ord` implementation at all. In my mind, forcing the caller to convert hashes to bytes to sort them has some value, because it makes it clear that the comparison semantics are the regular `[u8]` comparison semantics. Given that we do encourage users to rely on our constant-time `Eq` impl, I worry it would be confusing to expose an `Ord` impl without making a similar guarantee.

PR #277, opened 2022-12-04
Update lib.rs

lambda0xE:
Minor performance improvements.


oconnor663:
Looks like there's a syntax error there. You'll want to make sure that `cargo check` and `cargo test` are clean.

Apart from that, could you tell me about the goal of the change?

Issue #281, opened 2023-01-12
Warn on stderr when producing output that cannot be checked with --check


joveian:
Currently per #33 and the --check [spec](https://github.com/BLAKE3-team/BLAKE3/blob/master/b3sum/what_does_check_do.md) b3sum --check rejects checking files with the unicode replacement character.  However, there is no warning or error when b3sum without --check produces output that can't be checked with --check, which seems less than ideal since checking files is usually the point of producing the list in the first place.  Since the spec seems to indicate that creating a list that can't be checked with --check is intended, I suggest a warning on stderr in that case.

joveian:
Sorry for the noise, I messed up using GitHub and should have had the changes on a branch so I could make another branch for other patches.

oconnor663:
> checking files is usually the point of producing the list in the first place

I'm not sure this is right. Personally, I never use `--check`, even though I use `b3sum` by hand frequently. My instinct is that we should optimize for the case where the human is reading `b3sum` output with their eyeballs.

That said, invalid unicode paths should also be pretty rare, so I don't think most people eyeball experience is going to be affected either way by what we decide here.

oconnor663:
What do you think of doing a summarized warning at the bottom. I think that might be less spammy, and also more likely to make the user see it, but possibly more confusing about which file caused the issue. Maybe something like this?

```diff
diff --git a/b3sum/src/main.rs b/b3sum/src/main.rs
index 91a87a1..31a3c15 100644
--- a/b3sum/src/main.rs
+++ b/b3sum/src/main.rs
@@ -565,6 +565,8 @@ fn main() -> Result<()> {
     let thread_pool = thread_pool_builder.build()?;
     thread_pool.install(|| {
         let mut some_file_failed = false;
+        let mut found_invalid_unicode = false;
+        let mut found_literal_replacement_char = false;
         // Note that file_args automatically includes `-` if nothing is given.
         for path in &args.file_args {
             if args.check() {
@@ -585,6 +587,26 @@ fn main() -> Result<()> {
                     eprintln!("{}: {}: {}", NAME, path.to_string_lossy(), e);
                 }
             }
+            match path.to_str() {
+                Some(s) => {
+                    if s.contains('�') {
+                        found_literal_replacement_char = true;
+                    }
+                }
+                None => found_invalid_unicode = true,
+            };
+        }
+        if found_invalid_unicode {
+            eprintln!(
+                "{}: a filename contains invalid unicode, output is not checkable",
[...]

PR #284, opened 2023-01-12
Warn of any unicode replacement characters

joveian:
Unicode replacement characters mean --check is not able to check the file so warn when generating them.  This time with NAME.

Fixes issue #281


joveian:
I guess the tests should be updated to capture and check stderr also, however I don't know rust and I can't figure out if the test is relying on the "error if not valid UTF-8" property of cmd.read or how to duplicate that property while capturing stderr as well if so.

PR #285, opened 2023-01-13
Add --check summary of failures

### LINKED ISSUE #283: b3sum --check does not produce a summary of failures
Unlike other similar tools, b3sum --check does not produce a summary of failures at the end of the output while also by default producing output for verified files.  This makes it easy to miss any failures.

joveian:
Add a warning to stderr at the end similar to other *sum utilities when --check is used and some files failed verification.

At least the last part of the comment above check_one_checkfile seemed to be incorrect so I removed that comment.

Fixes #283

### DIFF
--- a/b3sum/src/main.rs
+++ b/b3sum/src/main.rs
@@ -537,7 +537,7 @@ fn check_one_line(line: &str, args: &Args) -> bool {
     }
 }
 
-fn check_one_checkfile(path: &Path, args: &Args, files_failed: &mut i64) -> Result<()> {
+fn check_one_checkfile(path: &Path, args: &Args, files_failed: &mut u64) -> Result<()> {
     let checkfile_input = Input::open(path, args)?;
     let mut bufreader = io::BufReader::new(checkfile_input);
     let mut line = String::new();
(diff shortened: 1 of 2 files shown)


b3sum/src/main.rs
```diff
@@ -537,7 +537,7 @@ fn check_one_line(line: &str, args: &Args) -> bool {
     }
 }
 
-fn check_one_checkfile(path: &Path, args: &Args, some_file_failed: &mut bool) -> Result<()> {
+fn check_one_checkfile(path: &Path, args: &Args, files_failed: &mut i64) -> Result<()> {
```
oconnor663, line 540: Could we make this a `u64`? I'm thinking negative values wouldn't make sense here.

oconnor663:
Thank you! I added https://github.com/BLAKE3-team/BLAKE3/commit/f1dcbeadc238c940d0dcfa3608afee16c950685f and https://github.com/BLAKE3-team/BLAKE3/commit/4c51c06008d9338c3ea81f3ac031b81c0deead9b as followup tweaks.

PR #286, opened 2023-01-16
Use newer cc-rs to fix Mac Catalyst bug

imWildCat:


oconnor663:
My understanding is that, because these versions are compatible, the newer one should get used automatically without any change to Cargo.toml. Is that wrong?

imWildCat:
@oconnor663 Oh, sorry! You're right, there's no need to upgrade it in `Cargo.toml`.

```
[[package]]
name = "cc"
version = "1.0.78"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "a20104e2335ce8a659d6dd92a51a767a0c062599c73b343fd152cb401e828c3d"
```

It will be resolved to the latest compatible version.

PR #288, opened 2023-01-19
Make sign conversion explicit. Fix #287

### LINKED ISSUE #287: Sign conversion warning in C portable implementation.
I’m using the GCC option `-Wsign-conversion` to detect unintended sign conversions, and also `-Werror` and other checks. The functions `highest_one()` and `popcnt()` return `unsigned int`, but the compiler intrinsics [`__builitin_clzll()`](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#index-_005f_005fbuiltin_005fclzll) and [`__builtin_popcountll()`](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#index-_005f_005fbuiltin_005fpopcountll) return `int`:

```
git clone https://github.com/BLAKE3-team/BLAKE3.git
cd BLAKE3/c
gcc -Werror -Wsign-conversion -shared -O3 -o libblake3.so blake3.c blake3_dispatch.c blake3_portable.c
```
Output:
```
In file included from blake3.c:6:
blake3_impl.h: In function 'highest_one':
blake3_impl.h:90:13: error: conversion to 'unsigned int' from 'int' may change the sign of the result [-Werror=sign-conversion]
   90 |   return 63 ^ __builtin_clzll(x);
      |          ~~~^~~~~~~~~~~~~~~~~~~~
blake3_impl.h: In function 'popcnt':
blake3_impl.h:120:10: error: conversion to 'unsigned int' from 'int' may change the sign of the result [-Werror=sign-conversion]
  120 |   return __builtin_popcountll(x);
      |          ^~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

The fix is to make the conversion explicit by adding a cast to `unsigned int` when calling those functions in `blake3_impl.h`.

AlbertoGP:
Implicit sign conversions cause warnings when using -Wsign-conversion but that is easy to avoid by making the conversions explicit.

### DIFF
--- a/c/blake3_impl.h
+++ b/c/blake3_impl.h
@@ -87,7 +87,7 @@ static const uint8_t MSG_SCHEDULE[7][16] = {
 /* x is assumed to be nonzero.       */
 static unsigned int highest_one(uint64_t x) {
 #if defined(__GNUC__) || defined(__clang__)
-  return 63 ^ __builtin_clzll(x);
+  return 63 ^ (unsigned int)__builtin_clzll(x);
 #elif defined(_MSC_VER) && defined(IS_X86_64)
   unsigned long index;
   _BitScanReverse64(&index, x);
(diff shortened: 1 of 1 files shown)


oconnor663:
Thank you!

PR #289, opened 2023-01-23
Correct section names on Windows GNU assembly

namazso:
Currently this data is placed into ".rodata", which is an unknown section, therefore has read-write protections by default. After this change, it would be in .rdata, which is for read only data, and therefore has such protections set.


sneves:
Submitter is correct. Typically `.rdata` is the section name on Windows for constants, and that is the case on the .asms, but I though that the GNU toolchain also recognized `.rodata`. It doesn't.

oconnor663:
Thanks @namazso !

PR #293, opened 2023-02-17
Suppress MemorySanitizer false positive

jbms:
MemorySanitizer does not support assembly, and therefore produces a false positive on `blake3_hasher_finalize` and related functions.

Note that a similar (but less comprehensive) fix was previously applied to LLVM's fork of this repo:

https://github.com/llvm/llvm-project/blob/34aff47521c3e0cbac58b0d5793197f76a304295/llvm/lib/Support/BLAKE3/blake3.c#L576

### DIFF
--- a/c/blake3.c
+++ b/c/blake3.c
@@ -5,6 +5,13 @@
 #include "blake3.h"
 #include "blake3_impl.h"
 
+#ifdef __has_feature
+#if __has_feature(memory_sanitizer)
+#include <sanitizer/msan_interface.h>
+#define BLAKE3_MEMORY_SANITIZER_BUILD
+#endif
+#endif
+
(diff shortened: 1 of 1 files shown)


oconnor663:
Help me make sure I understand the problem. MSan thinks some memory is uninitialized, because the writes that initialized it were inside an assembly block/file that MSan can't analyze. Is that correct? That seems like it would be a problem for all sorts of libraries, like OpenSSL. Does OpenSSL include this sort of MSan hinting?

jbms:
Yes, that's all correct.

https://github.com/openssl/openssl/blob/6c0ecc2bce64cc86948a51f80f832b5e48a9ebea/crypto/engine/eng_rdrand.c#L47

PR #304, opened 2023-05-11
Refactor CMake buildsystem to be portable and modern

### LINKED ISSUE #102: add a CMake build system for the C code?
Is CMake the closest thing to a cross-platform standard for building C these days?

BurningEnlightenment:
This is building on top of @SteveGremory's #247
Resolves #102

Aggreggate source files directly in the target instead of a proxy variable.

Install CMake package config files in order to allow the project to be found via `find_package()` by dependents.

Remove hard coded compiler flags (including -fPIC). These are not portable and should be set by the toolchain file or on the CLI.

Replace hard coded SIMD compiler flags with configurable options. Retain the current GCC/Clang flags as defaults for these compilers. Add default SIMD compiler flags for MSVC.

Guard ASM sources with triplet compatibility checks.

Remove the `BLAKE3_STATIC` option in favor of [`BUILD_SHARED_LIBS`].

[`BUILD_SHARED_LIBS`]: https://cmake.org/cmake/help/v3.9/variable/BUILD_SHARED_LIBS.html

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -128,7 +128,7 @@ configure_package_config_file(blake3-config.cmake.in
 write_basic_package_version_file(
     "${CMAKE_CURRENT_BINARY_DIR}/blake3-config-version.cmake"
     VERSION ${libblake3_VERSION}
-    COMPATIBILITY SameMinorVersion # as long as libblake3_VERSION < 1.0.0
+    COMPATIBILITY SameMajorVersion
 )
 install(FILES
         "${CMAKE_CURRENT_BINARY_DIR}/blake3-config.cmake"
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -35,10 +35,19 @@ add_library(blake3
 add_library(BLAKE3::blake3 ALIAS blake3)
 
 # library configuration
+set(BLAKE3_PKGCONFIG_CFLAGS)
+if (BUILD_SHARED_LIBS)
+  target_compile_definitions(blake3 
+    PUBLIC BLAKE3_DLL
+    PRIVATE BLAKE3_DLL_EXPORTS
+  )
+  list(APPEND BLAKE3_PKGCONFIG_CFLAGS -DBLAKE3_DLL)
+endif()
 target_include_directories(blake3 PUBLIC $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 set_target_properties(blake3 PROPERTIES
   VERSION ${PROJECT_VERSION}
   SOVERSION 0
+  C_VISIBILITY_PRESET hidden
 )
 
 # optional SIMD sources
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -1,74 +1,132 @@
 cmake_minimum_required(VERSION 3.9)
 
-project(libblake3 VERSION 0.1.0 DESCRIPTION "BLAKE3 C implementation")
-enable_language(C ASM)
+project(libblake3
+  VERSION 0.1.0
+  DESCRIPTION "BLAKE3 C implementation"
+  LANGUAGES C ASM
+)
 
 include(GNUInstallDirs)
 
-option(BLAKE3_STATIC ON)
+# default SIMD compiler flag configuration (can be overriden by toolchains or CLI)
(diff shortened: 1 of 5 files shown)


c/CMakeLists.txt
```diff
@@ -1,74 +1,141 @@
 cmake_minimum_required(VERSION 3.9)
 
-project(libblake3 VERSION 0.1.0 DESCRIPTION "BLAKE3 C implementation")
-enable_language(C ASM)
+project(libblake3
+  VERSION 0.1.0
```
BurningEnlightenment, line 4:
```suggestion
  VERSION 1.3.3
```
I guess it would make sense to set this to `1.3.3` to match `BLAKE3_VERSION_STRING`, right?

c/CMakeLists.txt
```diff
@@ -1,74 +1,141 @@
...
+configure_package_config_file(blake3-config.cmake.in
+    "${CMAKE_CURRENT_BINARY_DIR}/blake3-config.cmake"
 
-endif(BLAKE3_STATIC)
+    INSTALL_DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/blake3"
+)
+write_basic_package_version_file(
+    "${CMAKE_CURRENT_BINARY_DIR}/blake3-config-version.cmake"
+    VERSION ${libblake3_VERSION}
+    COMPATIBILITY SameMinorVersion # as long as libblake3_VERSION < 1.0.0
```
BurningEnlightenment, line 131:
```suggestion
    COMPATIBILITY SameMajorVersion
```
If the previous suggestion wrt the version is accepted, this ought be the correct policy.

c/blake3.h
```diff
@@ -38,20 +60,20 @@ typedef struct {
...
+BLAKE3_API void blake3_hasher_init_derive_key(blake3_hasher *self, const char *context);
+BLAKE3_API void blake3_hasher_init_derive_key_raw(blake3_hasher *self, const void *context,
+                                                  size_t context_len);
+BLAKE3_API void blake3_hasher_update(blake3_hasher *self, const void *input,
+                                     size_t input_len);
+BLAKE3_API void blake3_hasher_finalize(const blake3_hasher *self, uint8_t *out,
+                                       size_t out_len);
+BLAKE3_API void blake3_hasher_finalize_seek(const blake3_hasher *self, uint64_t seek,
+                                            uint8_t *out, size_t out_len);
+BLAKE3_API void blake3_hasher_reset(blake3_hasher *self);
```
oconnor663, line 76: Could you say more about what these macros are doing? I'm a little nervous about leaking complexity into the header files that might affect folks who aren't using CMake to build this.

oconnor663:
Thanks for putting this together! I'm going to learn a lot reading through it :)

c/blake3.h
BurningEnlightenment: These macros control which symbols (=> C functions / global variables) end up in the shared object's (.so) symbol table, i.e. what you can access via [`dlsym()`](https://man7.org/linux/man-pages/man3/dlsym.3.html). The default on gcc and clang is to export all entities with external linkage while the Windows / MSVC defaults to exporting only entities annotated with `__declspec(dllexport)`. Without these annotations you basically can't use ther blake3.dll. So these aren't actually CMake specific. They are useful regardless of the buildsystem. You can see a similar macro used by libb2 here:
https://github.com/BLAKE2/libb2/blame/6fafe5c546860ecbdd2b97093a93ae4ce54dfe52/src/blake2.h#L20-L44

You can read more about symbol visibility on the [GCC Wiki](https://gcc.gnu.org/wiki/Visibility) and the [MSVC docs](https://learn.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-declspec-dllexport?view=msvc-170).

Also note that I configured CMake to set the default visibility to hidden for GCC / clang.

BurningEnlightenment:
@oconnor663 do you have further questions or can we get this merged?

oconnor663:
Merged! Thank you! I'll admit I still have no idea what I'm doing, and I only kicked the tires a little bit, but it makes me really happy that this works:

```
cmake .
make
gcc example.c libblake.a
```

I guess the next question is how we should document this? :)

BurningEnlightenment:
I can write a section for the `c/README.md` tomorrow or so.

On a related note I'm going to update the vcpkg port based on this; maybe someone else does the work for conan.

PR #307, opened 2023-05-24
build(cmake): Correctly detect x86 and arm64 Windows

BurningEnlightenment:
@oconnor663 I hate to bother you again, but microsoft/vcpkg#31602 revealed that I missed a few `CMAKE_SYSTEM_PROCESSOR` values for `x86` and `arm64`. Furthermore I added escape hatches just in case that there are even more synonyms for these ISAs. See the commit messages for further details.

I took this as an opportunity to log the active SIMD configuration.


oconnor663:
Thank you for putting all this effort in, and especially for the vcpkg.

PR #314, opened 2023-06-12
[CMake] Rework NEON detection

### LINKED ISSUE #312: aarch64: gcc: error: unrecognized command-line option '-mfpu=neon'
```
+ cmake  ...
...
* NEON SIMD intrinsics, The library uses NEON SIMD intrinsics.
...
[00:00:05] [ 80%] Building C object CMakeFiles/blake3.dir/blake3_neon.c.o
[00:00:05] /usr/bin/cc -DBLAKE3_DLL -DBLAKE3_DLL_EXPORTS -Dblake3_EXPORTS  -pipe -frecord-gcc-switches -Wall -g -O2 -flto=auto -fPIC -fvisibility=hidden -mfpu=neon -MD -MT CMakeFiles/blake3.dir/blake3_neon.c.o -MF CMakeFiles/blake3.dir/blake3_neon.c.o.d -o CMakeFiles/blake3.dir/blake3_neon.c.o -c /usr/src/RPM/BUILD/libblake3-1.4.0/c/blake3_neon.c
[00:00:05] aarch64-alt-linux-gcc: error: unrecognized command-line option '-mfpu=neon'
```
It seems GCC does not support `-mfpu=neon` flag on aarch64 (but supports this flag on armv7).

You may test this in Docker (which have qemu-user support) for example:
```
$ docker run --rm -it arm64v8/ubuntu:lunar
root@f3ff70e26edf:/# apt-get update
root@f3ff70e26edf:/# apt-get install -y gcc
root@f3ff70e26edf:/# gcc -mfpu=neon
gcc: error: unrecognized command-line option '-mfpu=neon'
gcc: fatal error: no input files
compilation terminated.
```

BurningEnlightenment:
Given the myriad of `-mfpu` options for ARM [[1]], the inability to portably query for CPU support, and the lack of standardized ISA names we have no other choice, but to opt out of automatically supplying NEON compile flags. Instead we simply add the NEON optimized source file if we detect an ISA with guaranteed NEON support (>= ARMv8) or the user explicitly requests it (in which case he is expected to provide the compile flags with `CMAKE_C_FLAGS` or `BLAKE3_CFLAGS_NEON` either through a toolchain file or commandline parameters).

Resolves #312 

[1]: https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -24,12 +24,11 @@ elseif(CMAKE_C_COMPILER_ID STREQUAL "GNU"
   set(BLAKE3_CFLAGS_SSE4.1 "-msse4.1" CACHE STRING "the compiler flags to enable SSE4.1")
   set(BLAKE3_CFLAGS_AVX2 "-mavx2" CACHE STRING "the compiler flags to enable AVX2")
   set(BLAKE3_CFLAGS_AVX512 "-mavx512f -mavx512vl" CACHE STRING "the compiler flags to enable AVX512")
-  set(BLAKE3_CFLAGS_NEON "-mfpu=neon" CACHE STRING "the compiler flags to enable NEON")
 endif()
 # architecture lists for which to enable assembly / SIMD sources
 set(BLAKE3_AMD64_NAMES amd64 AMD64 x86_64)
 set(BLAKE3_X86_NAMES i686 x86 X86)
-set(BLAKE3_ARM_NAMES aarch64 AArch64 arm64 ARM64 armv8 armv8a)
+set(BLAKE3_ARMv8_NAMES aarch64 AArch64 arm64 ARM64 armv8 armv8a)
 
 # library target
 add_library(blake3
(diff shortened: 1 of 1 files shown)


oconnor663:
This sounds similar to what we do in the Rust build, so it sounds like a good change to me.

Is there a reason we need both `BLAKE3_USE_NEON` and `BLAKE3_USE_NEON_INTRINSICS`? Could we unify them into just `BLAKE3_USE_NEON`?

BurningEnlightenment:
> Is there a reason we need both BLAKE3_USE_NEON and BLAKE3_USE_NEON_INTRINSICS? Could we unify them into just BLAKE3_USE_NEON, which is already in use?

The former is a C define, the latter is a CMake variable which is _requesting_ to build with NEON intrinsic support (which may not be granted).

oconnor663:
The approach we've been using in other places is that the default is "try to be smart about using NEON" (mainly just use it on AArch64 I think), and then if `BLAKE3_USE_NEON` is set we interpret `1` as "always use NEON (and possibly trigger compiler errors but so be it)" or `0` as "never use NEON". I like that that keeps the behavior of our explicitly flags as simple as possible, because we're probably on the hook for keeping those vaguely stable :)

That said, I'm going to land this as-is, and I'll let you make the call on whether we should make any more changes.

Issue #155, opened 2021-02-21
Usage of b3sum in the Rust as lib


ZeusMods:
Hello, is there a way to use b3sum as lib for my project?My project is also written in Rust, but I wonder here is how can I use b3sum as lib

So my Rust program will contain a list of prehashed files.
I want to call the b3sum to check these prehashed files list in file foo.blake3

oconnor663:
Right now `b3sum` doesn't expose a library interface of its own. But, have you considered using the `blake3` crate directly?

ZeusMods:
> Right now `b3sum` doesn't expose a library interface of its own. But, have you considered using the `blake3` crate directly?

Hello, it's pretty hard for me to work with blake3.

bronson:
Curious, what's the difficulty? That the crate works on buffers and not directly on files?

Banyc:
For people searching for a quick file hasher implementation:

```rust
fn file_hash<P>(file_path: P) -> io::Result<String>
where
    P: AsRef<Path>,
{
    let mut binary = std::fs::File::open(file_path)?;
    let mut reader = std::io::BufReader::new(&mut binary);
    let mut hasher = blake3::Hasher::new();
    while !reader.fill_buf()?.is_empty() {
        hasher.update(reader.buffer());
        reader.consume(reader.buffer().len());
    }
    let hash = hasher.finalize();
    Ok(hash.to_string())
}
```

oconnor663:
Now that I'm thinking about this again, it does seem like it might be nice to expose `b3sum`'s "memory map and call `update_rayon` for files bigger than 16 KiB" strategy. It's not very much code, but memory mapping isn't a familiar topic for most people, and it also involves a little bit of `unsafe`. It could be a `hash_path_maybe_mmap` function, something like this:

```rust
use std::fs::File;
use std::io::{self, prelude::*};
use std::path::Path;

// A 16 KiB buffer is enough to take advantage of all the SIMD instruction sets that we support,
// but `std::io::copy` currently uses 8 KiB. Most platforms can support at least 64 KiB on the
// stack, and there's some performance benefit to using bigger reads, so that's what we use here.
fn copy_wide(mut reader: impl Read, hasher: &mut blake3::Hasher) -> io::Result<u64> {
    let mut buffer = [0; 65536];
    let mut total = 0;
    loop {
        match reader.read(&mut buffer) {
            Ok(0) => return Ok(total),
            Ok(n) => {
[...]

PR #326, opened 2023-07-17
Move file operations from `b3sum` to `blake3`

Banyc:
Previously, the "memory map and call update_rayon for files bigger than 16 KiB" strategy is implemented in `b3sum` crate, which is not available for being a library. After the change, `blake3` users can now call `hash_path_maybe_mmap` to speed up file hashing.

This PR is related to #155.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,6 +79,8 @@ no_neon = []
 
 zeroize = ["zeroize_crate", "arrayvec/zeroize"]
 
+file = ["memmap2", "rayon", "std"]
+
 [package.metadata.docs.rs]
 # Document Hasher::update_rayon on docs.rs.
 features = ["rayon"]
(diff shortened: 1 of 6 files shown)


Banyc:
`copy_wide` is put at `lib.rs` right under `Hasher` `impl`'s because it can be used other than file operations. `copy_wide` and `maybe_memmap_file` become public because `b3sum` needs to access them directly.

oconnor663-travel:
Apologies for not responding to this PR more promptly. I'm out of the country and mostly off the grid until August 20.

oconnor663:
I've made several commits on top of this one and opened another PR at https://github.com/BLAKE3-team/BLAKE3/pull/342. I'd love to get your thoughts.

oconnor663:
Landed as e0bb91564125407102af81e219399025aa2c24b9 and cb32f0bd1450991d86a8ceb3716fd591382cc507.

oconnor663:
Just released as 1.5.0 :) That last commit got rid of the separate module entirely, in favor of inherent methods on `Hasher`.

PR #329, opened 2023-07-29
Add function to efficiently hash entire non-root subtrees to guts

rklaehn:
provides an efficient way to compute the hash for multiple chunks that do not start at chunk 0, as part of a larger tree.

This is useful for crates like [bao](https://github.com/oconnor663/bao), [abao](https://github.com/n0-computer/abao) and [bao-tree](https://github.com/n0-computer/bao-tree) when using chunk groups.

Adding this required to disable some assertions that contained assumptions that are no longer true.

This gives a performance improvement of 4x when computing an outboard with a chunk group size of 16kb in bao-tree compared to using `blake3::guts::ChunkState` and `blake3::guts::parent_cv` to compute a hash for a chunk group.

UPDATE:

this works in the case of bao-tree, but the signature is not good since it can't work in the general case. E.g.

```rust
let data = [0u8; 2048];
let hash = hash_block(1, &data, false);
```

This does not make any sense at all, since there is no single root for chunks 1 and 2. I guess you could put in an assert to catch things like this?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -409,6 +409,18 @@ impl Output {
         Hash(platform::le_bytes_from_words_32(&cv))
     }
 
+    fn hash(&self, is_root: bool) -> Hash {
+        // debug_assert_eq!(self.counter, 0);
+        let mut cv = self.input_chaining_value;
+        let mut flags = self.flags;
+        if is_root {
+            flags |= ROOT;
+        }
+        self.platform
+            .compress_in_place(&mut cv, &self.block, self.block_len, 0, flags);
+        Hash(platform::le_bytes_from_words_32(&cv))
+    }
+
     fn root_output_block(&self) -> [u8; 2 * OUT_LEN] {
         self.platform.compress_xof(
             &self.input_chaining_value,
@@ -977,6 +989,15 @@ impl Hasher {
         Self::new_internal(IV, 0)
     }
 
+    /// Construct a new `Hasher` with a start chunk
+    fn new_with_start_chunk(start_chunk: u64) -> Self {
+        Self {
+            key: *IV,
+            chunk_state: ChunkState::new(IV, start_chunk, 0, Platform::detect()),
+            cv_stack: ArrayVec::new(),
+        }
+    }
+
     /// Construct a new `Hasher` for the keyed hash function. See
     /// [`keyed_hash`].
     ///
@@ -1246,7 +1267,7 @@ impl Hasher {
         // also. Convert it directly into an Output. Otherwise, we need to
         // merge subtrees below.
         if self.cv_stack.is_empty() {
-            debug_assert_eq!(self.chunk_state.chunk_counter, 0);
+            // debug_assert_eq!(self.chunk_state.chunk_counter, 0);
             return self.chunk_state.output();
         }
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1246,7 +1267,7 @@ impl Hasher {
         // also. Convert it directly into an Output. Otherwise, we need to
         // merge subtrees below.
         if self.cv_stack.is_empty() {
-            debug_assert_eq!(self.chunk_state.chunk_counter, 0);
+            // debug_assert_eq!(self.chunk_state.chunk_counter, 0);
```
rklaehn, line 1270: This must be disabled because the chunk counter can be set to a non zero value in the constructor now.

src/lib.rs
```diff
@@ -1265,11 +1286,11 @@ impl Hasher {
         let mut output: Output;
         let mut num_cvs_remaining = self.cv_stack.len();
         if self.chunk_state.len() > 0 {
-            debug_assert_eq!(
-                self.cv_stack.len(),
-                self.chunk_state.chunk_counter.count_ones() as usize,
-                "cv stack does not need a merge"
-            );
+            // debug_assert_eq!(
```
rklaehn, line 1289: This must be disabled since because the chunk counter does not necessarily start at 0, it's current value is no longer equivalent to the number of chunks.

I don't think there is a way to reenable this except by adding an additional value to the state that tracks the start chunk.

src/lib.rs
```diff
@@ -1304,6 +1325,11 @@ impl Hasher {
         self.final_output().root_hash()
     }
 
+    fn finalize_node(&self, is_root: bool) -> Hash {
```
rklaehn, line 1328: Just exposes a way to create non root hashes

src/lib.rs
```diff
@@ -409,6 +409,18 @@ impl Output {
         Hash(platform::le_bytes_from_words_32(&cv))
     }
 
+    fn hash(&self, is_root: bool) -> Hash {
```
rklaehn, line 412: Just exposes a way to create non root hashes

rklaehn:
I am not quite sure, but it seems that what this blake3 fork is doing would also be covered with this: https://github.com/fleek-network/BLAKE3/commit/0e15418f195d138a52f5b86210d721fcb4f370d5

oconnor663-travel:
I'm out of the country and mostly offline until August 20, so I won't be able to take a proper look until then. But you might be interested in the new "guts" API I'm in the middle of:

- https://github.com/BLAKE3-team/BLAKE3/tree/guts_api/rust/guts
- https://github.com/oconnor663-travel/BLAKE3/tree/guts_api/rust/guts

Apologies for the big pile of undocumented code there :) But my hope is that I can port BLAKE3's existing SIMD implementations to that expanded API and then release that as a (permanently unstable) crate that other low-level callers can depend on. Here are a couple other projects I hope to support with the same API:

- https://github.com/oconnor663/bessie
- https://github.com/oconnor663/blake3_aead

rklaehn:
@oconnor663-travel this looks awesome. No stress, we have forked the crate and are depending on that one now, so there is not much urgency. Once a crate is out that allows us to compute subtree hashes efficiently, we will yank that one and go back to upstream.

Happy to close this if there is something better in the works. It would be nice if there was an example or a fn for the hash_subtree use case implemented in this PR.

Enjoy being offline.

PR #331, opened 2023-08-16
Fix Windows build with clang-cl

rui314:
clang-cl is LLVM's MSVC-compatible compiler frontend for Windows ABI. If clang-cl is in use, `CMAKE_C_COMPILER_ID` is `Clang` even though it doesn't take Unix-like command line options but MSVC-like options.

`if(MSVC)` is the correct predicate to check if we should pass MSVC-ish command line options.


BurningEnlightenment (approved):
@oconnor663 LGTM

PR #332, opened 2023-08-24
Fix CMake target include directories

r4nx:
Include directory is not set if the library is used with `add_subdirectory` or `FetchContent`, this PR fixes it


c/CMakeLists.txt
```diff
@@ -47,7 +47,10 @@ if (BUILD_SHARED_LIBS)
   )
   list(APPEND BLAKE3_PKGCONFIG_CFLAGS -DBLAKE3_DLL)
 endif()
-target_include_directories(blake3 PUBLIC $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
+target_include_directories(blake3 PUBLIC
+  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
```
BurningEnlightenment, line 51:
```suggestion
  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
```
Less stuff to change if we ever split the `CMakeLists.txt`.

BurningEnlightenment (changes requested):
Apart from the suggestion I'm okay with this.

r4nx:
> Apart from the suggestion I'm okay with this.

Thanks, sure, I've commited the changes

BurningEnlightenment (approved):
@oconnor663 LGTM

PR #335, opened 2023-08-30
Fix build for big endian AArch64 hosts

kovdan01:
The implementation does not support using arm neon on big-endian hosts: see blake3_neon.c. Setting `BLAKE3_USE_NEON` to 1 by default for all AArch64 hosts broke builds for big endian hosts. This patch fixes the behavior by introducing an additional check against `__ARM_BIG_ENDIAN` before setting `BLAKE3_USE_NEON`.

### DIFF
--- a/c/blake3_impl.h
+++ b/c/blake3_impl.h
@@ -49,8 +49,9 @@ enum blake3_flags {
 #endif
 
 #if !defined(BLAKE3_USE_NEON) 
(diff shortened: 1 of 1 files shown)


oconnor663:
I happened to get to https://github.com/BLAKE3-team/BLAKE3/pull/280 before this PR, but feel to rebase this and make it cleanup if you like.

Issue #290, opened 2023-01-25
Provide optional serde support for blake3::Hash


rklaehn:
It would be convenient for blake3::Hash to have optional serde support.

I checked if other hashing libs do this. E.g. the sha2 lib directly produces `GenericArray`, which does support serde.

So - it is not the end of the world if you don't want to do it, but it would be nice and very straightforward to implement.

oconnor663:
This also came up in https://github.com/BLAKE3-team/BLAKE3/pull/147. I'm kind of 50/50 on the idea. A couple of disorganized thoughts:

- I've been thinking of factoring out the `Hash` type into its own separate library, which could be shared with other crates. It already supports constant-time comparison and hex encoding/decoding, which is code that I find myself copy-pasting to other hash implementations I maintain. Maybe with a factored-out implementation, it would make more sense to add features like this.
- Is there a way to control whether serde serializes things as a list of ints vs as a hex string? I imagine some users would want one, and other users would want the other.

rklaehn:
So this external hash crate would be not just for blake3 hashes but for generic hashes, and provide things like constant-time ops that are relevant for hashes? That would be nice. But then would you have to wrap it into a newtype to prevent mixing different kinds of hashes?

Regarding the serde thing. Not sure what is the best way to do it. But there is a mechanism in serde to detect whether a format intends to be human readable. Maybe use that to detect whether to use blob or hex string?

https://docs.rs/serde/latest/serde/trait.Deserializer.html#method.is_human_readable

qti3e:
This code which I've written before exactly does this, and I don't think it needs any modification and can be copy-pasted into the codebase and activated with a `serde` feature flag. This uses hex encoding for human-readable formats (such as JSON and toml) and a fixed-size binary slice in binary formats (such as barcode, cbor, etc.)

See the test cases at the end of the same file.

https://github.com/qti3e/ross/blob/6cfb3502de90bca4a1444dd43a1d746c9a49060a/ross_core/src/utils/hash.rs#L84-L179

Let me know if you want this as a feature, and I'll happily open a pull request.

oconnor663:
I've cobbled together what I think a generic, Serde-supporting `Hash` type might look like: https://github.com/oconnor663/cryptographic_hash. Unfortunately, it looks like the implementation will be pretty awkward without some more const generics features in stable. Curious to get other people's thoughts.

rminderhoud:
[...]

PR #339, opened 2023-09-09
Add serde support for Hash behind optional feature

rminderhoud:
Added a new cargo feature `serde` that when enabled will derive `serde::Serialize` and `serde::Deserialize` for the `blake3::Hash` struct.

Fixes https://github.com/BLAKE3-team/BLAKE3/issues/290

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,11 +95,12 @@ features = ["mmap", "rayon", "zeroize"]
 arrayref = "0.3.5"
 arrayvec = { version = "0.7.4", default-features = false }
 constant_time_eq = "0.3.0"
-rayon = { version = "1.2.1", optional = true }
 cfg-if = "1.0.0"
 digest = { version = "0.10.1", features = [ "mac" ], optional = true }
-zeroize = { version = "1", default-features = false, features = ["zeroize_derive"], optional = true }
 memmap2 = { version = "0.7.1", optional = true }
+rayon = { version = "1.2.1", optional = true }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -79,6 +79,9 @@ no_neon = []
 
 zeroize = ["zeroize_crate", "arrayvec/zeroize"]
 
+# Enable optional serialization and deserialization of the Hash type using serde
+serde = ["dep:serde"]
```
oconnor663, line 83: This is equivalent to the default behavior if we just omit this line, right? If we want to keep it, we should move it above the comment that says "Features below this line are undocumented and unstable".

oconnor663:
Lots of folks have asked for this over the last couple years, so it probably does make sense to go ahead and include it. Thanks for opening the PR!

Cargo.toml
rminderhoud: That's a nice trick, didn't know about it. Made the change as requested (line removed)

rminderhoud:
> Lots of folks have asked for this over the last couple years, so it probably does make sense to go ahead and include it. Thanks for opening the PR!

Thanks, I'm glad you are willing to merge it into main!

oconnor663:
Could you rebase on the most recent master? There has been some churn in Cargo.toml.

rminderhoud:
Sure no problem, rebased with latest master branch

oconnor663:
Landed, thanks!

PR #341, opened 2023-09-10
Wasm32 SIMD Rust implementation

### LINKED ISSUE #187: add an implementation using Wasm SIMD
Rust just shipped version 1.54, which stabilizes Wasm SIMD intrinsics. We should port the SSE4.1 implementation to use these. First I should add Wasm to CI and make sure benchmarks work there.

monoid:
This code is based on `rust_sse2.rs` of the same distribution, and is subject to further improvements. Some comments are left intact even if their applicability is questionable.

SIMD implementation is gated by `wasm32_simd` feature, portable version is used otherwise.

Performance measurements with a primitive benchmark with ~16Kb of data:
```
| M1 native     | 11,610 ns |
| M1 WASM SIMD  | 13,355 ns |
| M1 WASM       | 22,037 ns |
| x64 native    |  6,713 ns |
| x64 intrins.  |  7,979 ns |
| x64 WASM SIMD | 11,985 ns |
| x64 WASM      | 25,978 ns |
```
wasmtime v12.0.1 was used on both platforms.

Closes #187.


monoid:
README should be updated too.

src/platform.rs
```diff
@@ -320,6 +358,12 @@ impl Platform {
         // Assumed to be safe if the "neon" feature is on.
         Some(Self::NEON)
     }
+
+    #[cfg(blake3_wasm32_simd)]
+    pub fn wasm32_simd() -> Option<Self> {
+        // Assumed to be safe if the "neon" feature is on.
```
oconnor663, line 364: paste-o in this comment

src/wasm32_simd.rs
```diff
@@ -0,0 +1,787 @@
...
+
+#[inline(always)]
+unsafe fn storeu(src: v128, dest: *mut u8) {
+    // This is an unaligned store, so the pointer cast is allowed.
+    v128_store(dest as *mut v128, src)
+}
+
+#[inline(always)]
+unsafe fn add(a: v128, b: v128) -> v128 {
+    i32x4_add(a, b)
```
oconnor663, line 41: The x86 versions of these intrinsics are all unsafe, I think because the machine opcodes they represent did undocumented stuff (besides just crashing) on very old Intel processors. But if I'm reading the docs correctly, the Wasm versions are safe, and a lot of these `unsafe` keywords can be removed.

Cargo.toml
```diff
@@ -94,7 +99,7 @@ zeroize_crate = { package = "zeroize", version = "1", default-features = false,
 
 [dev-dependencies]
 hex = "0.4.2"
-page_size = "0.5.0"
+page_size = "0.6.0"
```
oconnor663, line 102: I just pushed d6265dafc9366d59de273a831339dafe0058ceb7 for this, so it should rebase out.

src/platform.rs
```diff
@@ -12,6 +12,8 @@ cfg_if::cfg_if! {
         }
     } else if #[cfg(blake3_neon)] {
         pub const MAX_SIMD_DEGREE: usize = 4;
+    } else if #[cfg(blake3_wasm32_simd)] {
+        pub const MAX_SIMD_DEGREE: usize = 8;
```
oconnor663, line 16: Should this be 4? Either way, `MAX_SIMD_DEGREE` and `MAX_SIMD_DEGREE_OR_2` should definitely be equal if the former is 2 or greater.

src/platform.rs
```diff
@@ -97,6 +108,9 @@ impl Platform {
             Platform::AVX512 => 16,
             #[cfg(blake3_neon)]
             Platform::NEON => 4,
+            #[cfg(blake3_wasm32_simd)]
+            // TODO is it 8 or 4??? SSE4 has 4...
```
oconnor663, line 112: My understanding is that Wasm SIMD always uses 128-bit vectors, which would imply degree 4 (128/32). Am I missing anything?

src/wasm32_simd.rs
```diff
@@ -0,0 +1,787 @@
...
+    flags: u8,
+) {
+    let [row0, row1, row2, row3] = compress_pre(cv, block, block_len, counter, flags);
+    // it stores in reversed order...
+    storeu(xor(row0, row2), cv.as_mut_ptr().add(0) as *mut u8);
+    storeu(xor(row1, row3), cv.as_mut_ptr().add(4) as *mut u8);
+}
+
+#[target_feature(enable = "simd128")]
+pub unsafe fn compress_xof(
```
oconnor663, line 373: Similar to above, the x86 version of these fundamentally can never be safe, because you must not execute new instructions on old chips. But it seems like the Wasm versions can be safe here, as long as the arguments are regular safe reference types and stuff, because they're guaranteed to crash cleanly if support isn't there?

src/platform.rs
```diff
@@ -163,6 +182,11 @@ impl Platform {
             // No NEON compress_xof() implementation yet.
             #[cfg(blake3_neon)]
             Platform::NEON => portable::compress_xof(cv, block, block_len, counter, flags),
+            #[cfg(blake3_wasm32_simd)]
+            // TODO Safe because compiled for wasm32
```
oconnor663, line 186: Similar to my comments around the intrinsics: Hopefully we can just make this safe to call? If so the comment should probably be something like "this will crash if the host doesn't support Wasm SIMD".

oconnor663 (changes requested):
Thank you for putting this together! This must've been a _lot_ of work. And the benchmarks look wonderful.

One high-level question we should think about is whether we want to add Wasm support to the C implementation at the same time we add it to the Rust implementation. On the one hand, so far we've always done them both at the same time. On the other hand, Wasm is a very different beast from other SIMD extensions, dealing with different levels of support across different versions of GCC/Clang/MSVC seems like it would be a huge pain. What are your thoughts?

oconnor663:
By the way, the MIPS build failure in CI is already fixed, and it should go away if you rebase.

src/platform.rs
monoid: Exactly, but I see that for SSE4 that has same vector size, `pub const MAX_SIMD_DEGREE: usize = 8;` is defined.

src/wasm32_simd.rs
monoid: Unfortunately, stores and loads are still unsafe, so this function is not safe.

monoid:
Well, it took about 4 evenings, given that I have very little experience with SIMD instructions beyond general idea.  I'm still surprised it works :)

I have very limited knowledge of C support in WASM land, but I've googled nothing except Emscripten compiler.  And it seems it can translate SSE4 intrinsics to WASM:
https://emscripten.org/docs/porting/simd.html

src/platform.rs
monoid: The only safety concern now is correct stores and loads.  I will think about formulating the comment.

monoid: I've made some functions safe with small unsafe blocks inside.  I haven't converted `hash_many` and related functions, though.

oconnor663:
Apologies for the delay in reviewing this PR. I just wanted to mention that I haven't forgotten about it :)

monoid:
@oconnor663 Would it be a Christmas gift?

elichai:
@oconnor663 friendly ping :)

monoid:
Meanwhile one may use https://crates.io/crates/fluence-blake3

monoid:
@oconnor663 Could you, please, review this PR?

oconnor663:
Apologies for letting this sit for so long. Is there way for me to run `cargo test` (or dare I dream, `cargo bench`) to exercise this implementation? I can do the CI integration myself once I know how to run tests by hand.

monoid:
@oconnor663 it seems both tests and benches can be run!  I've added `.cargo/config.toml` that defines `wasmtime` as a runner for `wasm32-wasip1` target, and now you can run, given you have `wasmtime` installed:
```
rustup target add --toolchain nightly wasm32-wasip1
cargo +nightly bench --target=wasm32-wasip1
cargo +nightly bench --target=wasm32-wasip1 --features wasm32_simd
cargo +nightly test --target=wasm32-wasip1
cargo +nightly test --target=wasm32-wasip1 --features wasm32_simd
```
Other WASM implementation may be used as well, just edit the `.cargo/config.toml` or define an environment variable `CARGO_TARGET_WASM32_WASIP1_RUNNER`.

oconnor663:
Holy shit :sweat_smile: 

```
$ cargo +nightly bench --target wasm32-wasip1 atonce_1024
...
test bench_atonce_1024_kib             ... bench:   3,826,985.45 ns/iter (+/- 49,234.86) = 273 MB/s

$ cargo +nightly bench --target wasm32-wasip1 --features wasm32_simd atonce_1024
...
test bench_atonce_1024_kib             ... bench:     639,369.00 ns/iter (+/- 10,935.28) = 1640 MB/s
```

oconnor663:
Landed! Thanks again for doing all this work. I'll release this as v1.7.0 today.

monoid:
@oconnor663 thank you very much! Great news for my resumé :D

BTW, I'm fuzzing the wasm-simd implementation with the native one.  No differences yet in 24h*core.

PR #342, opened 2023-09-11
`io` module

oconnor663:
This is a proposed `blake::io` API based on @Banyc's work in https://github.com/BLAKE3-team/BLAKE3/pull/326. We should definitely bikeshed it a bit before we land anything.


Banyc:
LGTM and I have learned a lot!

oconnor663:
cb32f0bd1450991d86a8ceb3716fd591382cc507

PR #343, opened 2023-09-11
Cmultithreading

pantoniou:
This PR provides multithreading for the C-implementation.
It is quite minimal, but it gets the job done of getting C to withing 1% of the Rust implementation.

### DIFF
--- a/c/.gitignore
+++ b/c/.gitignore
@@ -1,4 +1,5 @@
 blake3
 example
+example-mmap
 build/
 *.o
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -35,6 +35,7 @@ add_library(blake3
   blake3.c
   blake3_dispatch.c
   blake3_portable.c
+  blake3_thread.c
 )
 add_library(BLAKE3::blake3 ALIAS blake3)
 
--- a/c/Makefile.testing
+++ b/c/Makefile.testing
@@ -46,7 +46,7 @@ ifdef BLAKE3_NO_NEON
 EXTRAFLAGS += -DBLAKE3_USE_NEON=0
 endif
 
-all: blake3.c blake3_dispatch.c blake3_portable.c main.c $(TARGETS)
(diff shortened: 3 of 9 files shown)


oconnor663:
Thank you for doing all this work! It might be a week or two before I have enough time to review it properly, but I'll be very interested to understand how it all works. Is it architecturally similar to Rayon, or something totally different?

I should be honest that I don't know if we're going to want to maintain this much threading machinery inside the project. One thing I'll want to think about is whether we could expose APIs that would make it possible to implement something like this on top of our code.

pantoniou:
Don't mention it. I had a need for a fast hash and BLAKE3 fits the bill.

I am not familiar with Rayon but from what I can understand from the exposed API the threading method is exactly the same.
I.e. the threading wait model.

Well, I wouldn't worry about it too much, since I'm maintaining a full blown C project with all the bells and whistles of detecting compiler versions and blake3 backends already.

Let me know if there's something I can do to make things easy for you guys to make next BLAKE3 version easier to drop in my codebase.

For example I could parsel out the BLAKE3 support as a standalone library that people could use like normal (i.e. apt-get install libcblake3) but I don't know if that's something you would be fine with. There's the small nit that I'm not much of a windows person so maybe target Linux/MacOS/unices first.

nirs:
I think it will be more useful to provide a C API to the rust code so it can be used as a library instead of maintaining duplicate implementations.

BurningEnlightenment:
This is notably missing Windows support. Furthermore I find it very intrusive if libraries spin up their own thread pools as there are machines with _hundreds_ of cores out there and PIds are very much a limited resource, i.e. this can provoke system resource exhaustion events for some people.

lelik107:
@oconnor663 My only question here is - are you going to multithread the С code to the end of the World? Something is better than nothing! You could mark it as experimental and add OMP dependency and than much more people would test how it actually works or it's a bad idea. But there nothing even similar I see in the project.

PR #353, opened 2023-10-01
build(CMake): Apply PP definitions to all sources

### LINKED ISSUE #350: Assertion fails on `armv7` when built with `NEON`
`blake3` (`c` implementation) built for `linux/amrv7` with  `cmake` and`NEON` support enabled fails on following assert (when used in `mold`):
```
ld.mold: /home/users/builder/rpm/BUILD/BLAKE3-1.5.0/c/blake3.c:343: compress_subtree_to_parent_node: Assertion `num_cvs <= MAX_SIMD_DEGREE_OR_2' failed.
```

It appears to be caused by compilation definition `BLAKE3_USE_NEON` being defined only for `blake3_dispatch.c`:
https://github.com/BLAKE3-team/BLAKE3/blob/5aa53f07f7188a569cadfc5daf1522972d9a9630/c/CMakeLists.txt#L134-L134

But not on `blake3.c` which uses `blake3_impl.h` and causing `MAX_SIMD_DEGREE` mismatch:
https://github.com/BLAKE3-team/BLAKE3/blob/5aa53f07f7188a569cadfc5daf1522972d9a9630/c/blake3_impl.h#L51-L70

BurningEnlightenment:
Fixes #350

This doesn't setup CI for these cases, because I haven't had time to fiddle with cross compiling, sorry.

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -59,8 +59,12 @@ macro(BLAKE3_DISABLE_SIMD)
   set(BLAKE3_SIMD_AMD64_ASM OFF)
   set(BLAKE3_SIMD_X86_INTRINSICS OFF)
   set(BLAKE3_SIMD_NEON_INTRINSICS OFF)
-  set_source_files_properties(blake3_dispatch.c PROPERTIES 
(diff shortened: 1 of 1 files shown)


BurningEnlightenment:
@oconnor663 there you go :)

PR #356, opened 2023-10-09
Add support for rkyv

billyb2:
[rkyv](https://github.com/rkyv/rkyv/) is a zero copy serialization framework. I use it in a lot of personal projects, and it would be really useful to have support for serializing to rkyv built in, seeing as you do something similar for serde. If this pull request is unwelcome, please let me know and I'll close this.


src/lib.rs
```diff
@@ -331,6 +336,45 @@ impl PartialEq<[u8]> for Hash {
...
+    #[inline]
+    fn eq(&self, other: &Hash) -> bool {
+        constant_time_eq::constant_time_eq(&self.0, &other.0)
+    }
+}
+
+#[cfg(feature = "rkyv")]
+impl Eq for ArchivedHash {}
+
 impl fmt::Display for Hash {
```
monoid, line 378: Shouldn't `ArchivedHash` implement `Display` and `Debug` too?  Also, `Hash` derives `std::hash::Hash`.  Might be useful as well.

PR #359, opened 2023-10-31
build(CMake): Provide NEON cflags for ARMv8 32bit

BurningEnlightenment:
ARMv8 CPUs are guaranteed to support NEON instructions. However, for 32bit ARMv8 triplets GCC needs to explicitly be configured to enable NEON intrinsics.

Different approach to fix the issue of #358

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# CMakeLists.txt whitespace fixups
+3e14f865d30271c74fc68d417af488ea91b66d48
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -9,6 +9,10 @@ project(libblake3
 include(FeatureSummary)
 include(GNUInstallDirs)
 
+# architecture lists for which to enable assembly / SIMD sources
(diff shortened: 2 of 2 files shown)


rui314:
LGTM

BurningEnlightenment:
@rui314 thanks for the Bug Report.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# CMakeLists.txt whitespace fixups
+3e14f865d30271c74fc68d417af488ea91b66d48
```
oconnor663, line 2: The important parts look good to me, but I'm not sure it's worth having this file at the top level. I might be biased though, becuase I use fugitive.vim, which makes it easy to step across bulk commits with `~`. What do you think?

BurningEnlightenment: The GitHub Blame view respects this file, i.e. this is useful for anyone who just takes a glance at the code through the Web UI. Also VSCode has an inline line blame which becomes less useful if style commits aren't ignored.

And while it's usually easy to step through the latest few commits, it gets especially tedious after style changing commits layer up deep in the history.

PR #362, opened 2023-11-14
Add rand_core impls for OutputReader

Taywee:
I initially started writing a separate crate for this, but the implementation is so minimal, I think it makes sense to roll it into the core library here.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1643,3 +1643,42 @@ impl std::io::Seek for OutputReader {
         Ok(self.position())
     }
 }
+
+#[cfg(feature = "rand")]
+impl rand_core::SeedableRng for OutputReader {
+    type Seed = [u8; 32];
+
+    #[inline]
+    fn from_seed(seed: Self::Seed) -> Self {
+        let mut hasher = Hasher::new();
+        hasher.update(&seed);
+        hasher.finalize_xof()
+    }
+}
+
+#[cfg(feature = "rand")]
+impl rand_core::RngCore for OutputReader {
+    #[inline]
+    fn next_u32(&mut self) -> u32 {
+        rand_core::impls::next_u32_via_fill(self)
+    }
+
+    #[inline]
+    fn next_u64(&mut self) -> u64 {
+        rand_core::impls::next_u64_via_fill(self)
+    }
+
+    #[inline]
+    fn fill_bytes(&mut self, dest: &mut [u8]) {
+        self.fill(dest);
+    }
+
+    #[inline]
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), rand_core::Error> {
+        self.fill(dest);
+        Ok(())
+    }
+}
+
+#[cfg(feature = "rand")]
+impl rand_core::CryptoRng for OutputReader {}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,10 @@ mmap = ["std", "dep:memmap2"]
 # Implement the zeroize::Zeroize trait for types in this crate.
 zeroize = ["dep:zeroize", "arrayvec/zeroize"]
 
+# Implement the rand_core Rng traits for OutputReader, allowing using it as
+# a rand::Rng.
+rand = ["dep:rand_core"]
+
 # This crate implements traits from the RustCrypto project, exposed here as the
 # "traits-preview" feature. However, these traits aren't stable, and they're
 # expected to change in incompatible ways before they reach 1.0. For that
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -1643,3 +1643,42 @@ impl std::io::Seek for OutputReader {
...
+        hasher.update(&seed);
+        hasher.finalize_xof()
+    }
+}
+
+#[cfg(feature = "rand")]
+impl rand_core::RngCore for OutputReader {
+    #[inline]
+    fn next_u32(&mut self) -> u32 {
+        rand_core::impls::next_u32_via_fill(self)
```
oconnor663, line 1663: Hmmm, this is going to be somewhat inefficient, because OutputReader doesn't buffer partially consumed output blocks. Each block is 64 bytes, and we're only using 4 bytes here, so if you wind up looping on this method then your RNG throughput will be 1/16th what it should be.

We could add a 64-byte buffer to OutputReader, but that would only be a partial fix. Eventually (soon I swear!) I'm going to land the refactoring that extends SIMD optimizations to cover extended output. At that point the buffer size we need for optimal throughput will be even larger, and it'll vary by platform. I'm kind of torn about whether we should do this buffering in OutputReader itself. Thoughts?

Taywee: Hm, well, I do think that buffering inside OutputReader could prevent performance footguns like this, but it implements Read already, so most users who are using std could just wrap it in a BufferedReader for the same effect. Personally, I prefer modular solutions to loading a type with functionality it doesn't need. OutputReader already is documented to point out that it is not buffering, a note that I missed.

An easy fix for this case is to implement [BlockRngCore instead and offer a newtype wrapper around one of the BlockRng structures](https://docs.rs/rand_core/latest/rand_core/block/index.html). I'll poke at that later today.

oconnor663: Oh interesting. I'm still trying to wrap my head around how Rand does things, where there are multiple different traits you can implement to become an RNG, and you only need to pick one?

Generating a block of `u32` doesn't seem too bad to me. We could either use a bit of unsafe code to reinterpret the array as `&mut [u8]` to fill it, or pull in the `bytemuck` crate to do the same thing. That said, I don't love this behavior:

> If the requested byte length is not a multiple of 4, some bytes will be discarded.

On the one hand, probably no one expects reading 1 byte at a time to be the most efficient way to do anything. On the other hand, there's no _good reason_ we need to waste bytes here. Maybe a custom struct would be better after all?

src/lib.rs
```diff
@@ -1643,3 +1643,42 @@ impl std::io::Seek for OutputReader {
...
 }
+
+#[cfg(feature = "rand")]
+impl rand_core::SeedableRng for OutputReader {
+    type Seed = [u8; 32];
+
+    #[inline]
+    fn from_seed(seed: Self::Seed) -> Self {
+        let mut hasher = Hasher::new();
+        hasher.update(&seed);
```
oconnor663, line 1654: This is totally arbitrary and won't affect performance either way, but I have a slight preference for using `Hasher::new_keyed` with the seed as the key.

Taywee:
This has been largely rewritten.  When adding buffering, I realized that it made no sense to have the buffering logic in the Rng-specific type.  I've written a BufOutputReader type that handles the buffering and implemented the Rng traits for that.

src/lib.rs
Taywee: Generally, RngCore is the important bit.  `BlockRngCore` lets it be used to be wrapped by `BlockRng` to auto-implement `RngCore`.  `ReseedingRng` also depends on `BlockRngCore`, but it all revolves around `u32` and `u64` chunks for some reason.  I really don't get that aspect, to be honest.  It seems like generating random bytes is the obvious baseline, but maybe I'm missing something.

src/lib.rs
Taywee: I'm fine with that.  This one was just the easiest to quickly write tests to verify using `b3sum`.

Taywee: I've now switched this to use new_keyed, and updated the tests appropriately.

Taywee:
I'll also note that I have pushed this functionality as the external [`rand_blake3`](https://crates.io/crates/rand_blake3) crate, with the buffering extracted into the [`load-buffer`](https://crates.io/crates/load-buffer) crate.  If you think this is too much feature creep for the `blake3` crate itself, it won't hurt my feelings to just keep that functionality in those separate crates.

elichai:
@oconnor663 Any updates on this? I think that a buffering `OutputReader` can be a performance improvement on some use-cases (consuming small chunks of output data)

PR #365, opened 2023-12-02
build(CMake): Require C99 mode

### LINKED ISSUE #348: One comma too many
In blake3_impl.h

```
// internal flags
enum blake3_flags {
  CHUNK_START         = 1 << 0,
  CHUNK_END           = 1 << 1,
  PARENT              = 1 << 2,
  ROOT                = 1 << 3,
  KEYED_HASH          = 1 << 4,
  DERIVE_KEY_CONTEXT  = 1 << 5,
  DERIVE_KEY_MATERIAL = 1 << 6,
};
```

The very last one throw a warning   
Not a very big issue, of course

BurningEnlightenment:
Specify language requirement as a [compile-feature] and force compiler extensions off ensuring portability problems are detected early on. Note that we do not use the `C_STANDARD` property, because it doesn't propagate to dependent targets and would prohibit users from compiling their code base with consistent flags / language configuations if they were to target a newer C standard. Similarly we do not configure `C_STANDARD_REQUIRED` as [compile-feature]s do not interact with it--they are enforced regardless.

Closes #348

[compile-feature]: https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#compile-feature-requirements

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -1,4 +1,9 @@
-cmake_minimum_required(VERSION 3.9)
+cmake_minimum_required(VERSION 3.9 FATAL_ERROR)
+
+# respect C_EXTENSIONS OFF without explicitly setting C_STANDARD
+if (POLICY CMP0128)
+  cmake_policy(SET CMP0128 NEW)
+endif()
 
 project(libblake3
   VERSION 1.5.0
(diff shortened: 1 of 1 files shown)


oconnor663:
Thank you! Out of curiosity, if I had to figure out the right way to do something like this myself, where would I have found it? This all seems...very non-obvious...

BurningEnlightenment:
Well, the main problem with CMake is the amount of cruft accumulated until CMake 3 refocused on targets and their properties. So if you google something and the solution involves `set()`ing `CMAKE_` variables it is probably outdated. The reference documentation is fairly comprehensive, but it requires a bit of digging to grok the way it's structured.

A good starting point is the [Tutorial section] of the reference manual. In this case it would've already covered 90% of the topic. The unintuitive parts about compiler extensions are things I've learned the hard way--through user complaints about things blowing up in their face 🙃

[Tutorial Section]: https://cmake.org/cmake/help/latest/guide/tutorial/index.html#guide:CMake%20Tutorial

PR #370, opened 2024-01-10
Implement Ord and PartialOrd for Hash

DBLouis:
This removes the dependencies on constant_time_eq and replaces it with subtle.

 #202


joshtriplett:
As far as I can tell, this implementation of cmp and partial_cmp doesn't seem to allow for the possibility of equality.

DBLouis:
It should be good now, I think 🤔

PR #373, opened 2024-01-22
land the portable implementation of the new blake3_guts crate

oconnor663:
My [`guts_api` branch](https://github.com/BLAKE3-team/BLAKE3/tree/guts_api) has been around for a while, and it's time that I start figuring out how to land it. The first step will be to land the portable implementation by itself, without anything depending on it. This will get a lot of APIs in place, so that other work can happen in parallel:

- open a branch integrating the main `blake3` crate with `blake3_guts` (this might not land for a while)
- add RISCV support to guts
- add AVX-512 support to guts

The longer run questions will be things like "do we want to generate the different flavors of assembly from scripts?" and "how should we support [Bao](https://github.com/oconnor663/bao), [Bessie](https://github.com/oconnor663/bessie), and [BLAKE3-AEAD](https://github.com/oconnor663/blake3_aead)?" This PR is about making step-by-step progress instead of trying to answer all these questions as once.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,6 +131,17 @@ jobs:
       run: cargo test
       working-directory: ./reference_impl
 
+    # the new guts crate
+    - name: guts test
+      run: cargo test --all-features
+      working-directory: ./rust/guts
+    - name: guts no_std build
+      run: cargo build --no-default-features
+      working-directory: ./rust/guts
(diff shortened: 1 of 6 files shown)


oconnor663:
Refactor `blake3` PR: https://github.com/BLAKE3-team/BLAKE3/pull/374

oconnor663:
RISCV support PR: https://github.com/BLAKE3-team/BLAKE3/pull/375

PR #375, opened 2024-01-22
factor out RISCV support from the guts_api branch

oconnor663:
TODO: figure out what environment variable should enable this

### DIFF
--- a/rust/guts/Cargo.toml
+++ b/rust/guts/Cargo.toml
@@ -9,6 +9,9 @@ documentation = "https://docs.rs/blake3_guts"
 readme = "readme.md"
 edition = "2021"
 
+[dependencies]
+cfg-if = "1.0.0"
+
 [dev-dependencies]
 hex = "0.4.3"
 reference_impl = { path = "../../reference_impl" }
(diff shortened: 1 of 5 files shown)


silvanshade:
> TODO: figure out what environment variable should enable this

In the RVV support for the rust crypto packages that I've been working on, I've generally been following the scheme where first, the feature gate for `v` is used (e.g., `#[cfg(target_feature = "v")]`) and then, for the extensions which don't have corresponding Rust features yet (but are exposed to the LLVM code generator), I've been adding `cfg` flags that follow the naming scheme `target_feature_<ext>`.

The intention is that these cfg flags would be replaced by proper target_feature gates once the features are exposed to Rust.

If you followed that scheme here, you'd end up with something like this:
```rust
#[cfg(all(target_feature = "v", target_feature = "zbb", target_feature_zvbb))]
...
```

If you're trying to parse this through the environment variables, it looks like the variable `CARGO_CFG_TARGET_FEATURE` is set to a comma separated list of the target features enabled. You should be able to get general cfg flags from `CARGO_CFG_<cfg>`, so then `target_feature_zvbb` should be set as `CARGO_CFG_TARGET_FEATURE_ZVBB`. Referenced [here](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts).

PR #382, opened 2024-02-03
silence -Wunused-function on MacOS-arm64

divinity76:
the #endif was in the wrong place, causing the error /Users/runner/work/php-src/php-src/ext/hash/blake3/upstream_blake3/c/blake3_dispatch.c:112:5: error: unused function 'get_cpu_features' [-Werror,-Wunused-function]

full compiler log https://github.com/php/php-src/actions/runs/7762643678/job/21173438425?pr=13194

### DIFF
--- a/c/blake3_dispatch.c
+++ b/c/blake3_dispatch.c
@@ -86,7 +86,6 @@ static void cpuidex(uint32_t out[4], uint32_t id, uint32_t sid) {
 #endif
 }
 
-#endif
 
 enum cpu_feature {
   SSE2 = 1 << 0,
@@ -161,6 +160,7 @@ static
 #endif
   }
 }
+#endif
 
 void blake3_compress_in_place(uint32_t cv[8],
                               const uint8_t block[BLAKE3_BLOCK_LEN],


oconnor663:
It looks like this move makes `#if defined(IS_X86)` on line 119 redundant. The intention was to eventually use `get_cpu_features` on other platforms, though we haven't done that yet. Is there any more direct way to mark the function "don't warn if this is unused"? Or is that all ugly compiler specific macros that we don't want to touch?

divinity76:
C23 has a solution: https://en.cppreference.com/w/c/language/attributes/maybe_unused
but it's 9 years too early to require C23 compatibility.

Maybe make get_cpu_features() always static, and make a public non-static blake3_get_cpu_features() which is just a wrapper for get_cpu_features() ? that would get rid of the unused warning.

maybe some more thought should be put into exactly what the public function should return tho, idk

divinity76:
adding
```
#include <assert.h>
assert(1 || get_cpu_features());
```
should work and be optimized away..

divinity76:
a possible alternative solution: https://github.com/BLAKE3-team/BLAKE3/pull/383

(it may solve the problem for ARM specifically, but probably not for other non-x86-non-arm architectures)

divinity76:
Don't have to use this fix, but would be nice if /a/ fix could be added. At 1.5.5  it still causes headache for integration in project compiling for MacOS with -Wunused-function -Werror

oconnor663:
I can't imagine we're going to be adding runtime feature detection to any other platforms anytime soon, so I'm just going to take this. Apologies for letting it sit for so long.

PR #383, opened 2024-02-05
runtime neon detection

divinity76:
tested on Oracle Cloud's cheapest ARM VPS VM.Standard.A1.Flex, seems to work. 

possible alternative for https://github.com/BLAKE3-team/BLAKE3/pull/382

### DIFF
--- a/c/blake3_dispatch.c
+++ b/c/blake3_dispatch.c
@@ -96,6 +96,7 @@ enum cpu_feature {
   AVX2 = 1 << 4,
   AVX512F = 1 << 5,
   AVX512VL = 1 << 6,
+  ARM_NEON = 1 << 7,
   /* ... */
   UNDEFINED = 1 << 30
 };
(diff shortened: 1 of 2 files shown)


sneves:
If you go to the [generic AArch64 manual](https://developer.arm.com/documentation/ddi0595/2021-12/AArch64-Registers/ID-AA64PFR0-EL1--AArch64-Processor-Feature-Register-0?lang=en) instead of the Cortex-A75 one, a value of 0 means that AdvSIMD is present, a value of 1 means AdvSIMD+FP16 is present, and a value of 15 means no AdvSIMD. I was under the impression that AdvSIMD was a required feature of AArch64. Go figure.

divinity76:
@sneves  hmm... does that mean
```
    uint64_t id_aa64pfr0_el1;
    __asm__ ("mrs %0, ID_AA64PFR0_EL1" : "=r" (id_aa64pfr0_el1));
    const uint8_t AdvSIMD = (id_aa64pfr0_el1 >> 20) & (1<<0 | 1<<1 | 1<<2 | 1 << 3);
    if(AdvSIMD != 15) {
      features = ARM_NEON;
    } else {
      features = 0;
    }
```
is better?

BurningEnlightenment:
As far as I know _standard_ ARMv8 implementations are required to support NEON, so apart from some rare special purpose CPUs NEON is guaranteed to be available. I heavily doubt that such a special purpose CPU would be used to run off-the-mill binaries / these people would compile with NEON=off anyways. Therefore it looks a whole lot like checking for SSE2 on x86-64 to me.

jblazquez:
Note that the code in this PR will need minor changes now that #389 has been merged.

Details here: https://github.com/BLAKE3-team/BLAKE3/pull/389#issuecomment-2041255514

divinity76:
> As far as I know _standard_ ARMv8 implementations are required to support NEON, so apart from some rare special purpose CPUs NEON is guaranteed to be available. I heavily doubt that such a special purpose CPU would be used to run off-the-mill binaries / these people would compile with NEON=off anyways. Therefore it looks a whole lot like checking for SSE2 on x86-64 to me.

Sounds good to me. Closing.

PR #384, opened 2024-02-09
optimize neon loadu_128/storeu_128

divinity76:
vld1q_u8 and vst1q_u8 has no alignment requirements.

This improves performance on Oracle Cloud's VM.Standard.A1.Flex by 1.15% on a 16*1024 input, from 13920 nanoseconds down to 13800 nanoseconds (approx)

### DIFF
--- a/c/blake3_neon.c
+++ b/c/blake3_neon.c
@@ -9,13 +9,13 @@
 #endif
 
 INLINE uint32x4_t loadu_128(const uint8_t src[16]) {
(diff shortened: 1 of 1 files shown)


oconnor663:
I see a ~1% improvement on the Graviton2 CPU on my AWS instance too. Thanks!

oconnor663:
Released as part of v1.5.1.

divinity76:
I wonder if this might have made big endian work too 🤔

PR #391, opened 2024-04-24
Generate shell completions and man page for `b3sum` at build time

### LINKED ISSUE #190: Missing man pages
Distributions could benefit from the inclusion of some man pages.
For example, ArchLinux has recently adopted the package in its official repos, but man pages are currently lacking:
<https://man.archlinux.org/listing/community/b3sum/>

### LINKED ISSUE #235: Shell completions for `b3sum`
Shell completions can be generated by [`clap_complete`](https://crates.io/crates/clap_complete). Do you have any plans to provide this?

sorairolake:
With this change, the following files will be generated when building `b3sum`:

- A man page (`b3sum.1`).
- Shell completions files for Bash (`b3sum.bash`), Elvish (`b3sum.elv`), fish (`b3sum.fish`), PowerShell (`_b3sum.ps1`), and Zsh (`_b3sum`).

To find where these files are generated run the following command in `b3sum/`:

```sh
find ./target -path '*/b3sum-*/out' -type d
```

To use these, install them to the destination (e.g., `/usr/share/man/man1` or `/usr/share/bash-completion/completions`).

Also, I set `rust-version` to `Cargo.toml`. This is based on the value of `.github/workflows/ci.yml`. If there is no value, `1.56.0` is set based on `edition`.

This closes #190 and closes #235.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
           # https://github.com/rust-lang/libs-team/issues/72.
           # This test target is here so that we notice if we accidentally bump
           # the MSRV, but it's not a promise that we won't bump it.
-          "1.66.1",
+          "1.72.0",
         ]
 
     steps:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ license = "CC0-1.0 OR Apache-2.0"
 documentation = "https://docs.rs/blake3"
 readme = "README.md"
 edition = "2021"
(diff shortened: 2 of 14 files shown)


sorairolake:
Bump the MSRV of `blake3` to 1.72.0. This is because the MSRV of [`zeroize` v1.8.0](https://crates.io/crates/zeroize/1.8.0) is 1.72.0, and the CI will fail if the MSRV of `blake3` isn't bumped.

PR #397, opened 2024-05-31
multilicensing for LLVM

veorq:
as discussed @oconnor663


oconnor663:
Added some more changes on top and merged. Nightly CI failures are unrelated new warnings, will fill those in a followup.

PR #399, opened 2024-06-18
build(windows): add support for gnullvm targets

ognevny:
there are 2 types of non-msvc windows targets: *-pc-windows-gnu and *-pc-windows-gnullvm. gnullvm targets based on llvm-mingw (clang toolchain in fact)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -74,7 +74,7 @@ fn is_big_endian() -> bool {
     endianness() == "big"
 }
 
-// Windows targets may be using the MSVC toolchain or the GNU toolchain. The
+// Windows targets may be using the MSVC toolchain or the MinGW toolchain. The
 // right compiler flags to use depend on the toolchain. (And we don't want to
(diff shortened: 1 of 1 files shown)


oconnor663:
Today I learned! Could you add a comment to this PR to help me remember next time I look at this line of code? :)

ognevny:
> Could you add a comment to this PR to help me remember next time I look at this line of code? :)

yes, but when?

oconnor663:
Sorry I meant like an inline comment in `build.rs` right above your change, reminding me/everyone to think about gnullvm when we change this code.

ognevny:
> Sorry I meant like an inline comment in `build.rs` right above your change, reminding me/everyone to think about gnullvm when we change this code.

ok, will add tomorrow

ognevny:
sorry for delay, I forgot about it! added a comment

PR #400, opened 2024-06-18
Switch to rayon-core for faster build

silvanshade:
This reduces the build time (with `--timings`) for `blake3` from around 1.7s to 1.3s on a Zen4 7950x.

This helps for situations where one is trying to minimize build time (e.g., for `build-dependencies`).

There seems to be no real downside since the crate doesn't use anything not included in the core rayon crate.

Below I've included the results from `cargo build --timings`.

The improvement doesn't really show up for `b3sum` on my system since other dependencies take long enough that it masks the difference. But it's still probably better to switch them both and maybe in some scenario the difference would surface.

[b3sum-rayon-core.pdf](https://github.com/user-attachments/files/15892931/b3sum-rayon-core.pdf)
[b3sum-rayon.pdf](https://github.com/user-attachments/files/15892932/b3sum-rayon.pdf)
[blake3-rayon-core.pdf](https://github.com/user-attachments/files/15892933/blake3-rayon-core.pdf)
[blake3-rayon.pdf](https://github.com/user-attachments/files/15892934/blake3-rayon.pdf)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ std = []
 # `update_rayon` and (in combination with `mmap` below) `update_mmap_rayon`
 # methods, for multithreaded hashing. However, even if this feature is enabled,
 # all other APIs remain single-threaded.
-rayon = ["dep:rayon", "std"]
+rayon = ["dep:rayon-core", "std"]
 
 # The `mmap` feature (disabled by default, but enabled for docs.rs) adds the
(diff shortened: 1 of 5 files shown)


oconnor663:
Wow yeah, if I do `cargo clean && time cargo build` on my box, this change brings it down from ~2.0s to ~1.2s. That's a surprisingly big difference. Thanks!

PR #405, opened 2024-07-02
build(CMake): support running tests with ctest

matoro:
Adds the ability to run the test suite via `ctest`.  Works with both in-source and out-of-source builds.  Respects user flags, and builds SIMD into the tests corresponding to whatever was used to build the library.  Does not interfere with the ability to run tests the old way.

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -168,6 +168,17 @@ else()
   BLAKE3_DISABLE_SIMD()
 endif()
 
+# cmake test support
+if (BLAKE3_BUILD_TESTING)
+  find_package(Python3 REQUIRED)
+  get_target_property(BLAKE3_SOURCES blake3 SOURCES)
+  add_executable(blake3-testing ${BLAKE3_SOURCES} main.c)
+  set_property(TARGET blake3-testing PROPERTY OUTPUT_NAME blake3)
+  target_compile_definitions(blake3-testing PRIVATE BLAKE3_TESTING=1)
+  enable_testing()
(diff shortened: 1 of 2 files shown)


c/CMakeLists.txt
```diff
@@ -168,6 +168,14 @@ else()
   BLAKE3_DISABLE_SIMD()
 endif()
 
+# cmake test support
+get_target_property(BLAKE3_SOURCES blake3 SOURCES)
+add_executable(blake3-testing ${BLAKE3_SOURCES} main.c)
+set_property(TARGET blake3-testing PROPERTY OUTPUT_NAME blake3)
+target_compile_definitions(blake3-testing PRIVATE BLAKE3_TESTING=1)
+enable_testing()
+add_test(test_vectors "${CMAKE_CURRENT_SOURCE_DIR}/test.py" WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}")
```
BurningEnlightenment, line 177: I think we should `find_package(PYTHON)` and invoke the python executable with the script in order to allow the user to specify the python installation.

BurningEnlightenment (changes requested):
Building tests should always be optional. Please add an option `BUILD_TESTING` or `BLAKE3_BUILD_TESTING`, conditionally call `enable_testing()`, and conditionally build the test target.

I'll defer to @oconnor663 as to whether we want to integrate the test environment with CMake.

matoro:
Sure thing, both suggestions done.

arthurzam:
ping?

It would really help as at Gentoo Linux.

BurningEnlightenment (approved):
An `option(BLAKE3_BUILD_TESTING …` would be nice to have for the GUI people. However, I'm fine with this as is. Thanks for the contribution 🙏 

@oconnor663 still needs to decide, whether we want this or not.

BurningEnlightenment:
I believe this has been superseeded by the TBB PR

PR #411, opened 2024-07-08
Manually implement `Zeroize`

gardk:
Removes dependence on proc-macros when enabling the zeroize feature.

As the macros previously made sure that all fields were zeroed, I opted for destructuring `self` in the impls so that a compile error will be triggered if a field is added and not handled.

Used `as_mut_slice().zeroize()` on arrays even though they implement `Zeroize`. This is because the implementation on slices zeroes the whole slice then issues an `atomic_fence()` call, while the array implementation calls it for every element. Don't know and haven't tested if there's a difference in optimizations. Both implementations produce the same result anyway :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,7 +103,7 @@ digest = { version = "0.10.1", features = [ "mac" ], optional = true }
 memmap2 = { version = "0.9", optional = true }
 rayon-core = { version = "1.12.1", optional = true }
 serde = { version = "1.0", default-features = false, features = ["derive"], optional = true }
-zeroize = { version = "1", default-features = false, features = ["zeroize_derive"], optional = true }
+zeroize = { version = "1", default-features = false, optional = true }
 
 [dev-dependencies]
 hmac = "0.12.0"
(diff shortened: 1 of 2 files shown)


oconnor663:
Thank you! This seems reasonable. Could you help me address a few points?

- Is there a measurable clean build time difference before and after?
- Can you confirm that something fails (either the build or existing tests) if any of these impls is deleted?
- Could you add inline comments next to spots where you're doing destructuring, with a quick explanation of why you're doing it?
- Similarly, could you comment on the array vs slice distinction inline, so that future readers (i.e. future me) will know why the difference matters.

gardk:
Added comments.

Here's the difference in clean build time on my machine.

```txt
# Manual
Time (mean ± σ):      1.356 s ±  0.202 s    [User: 2.496 s, System: 0.591 s]
Range (min … max):    1.197 s …  1.806 s    10 runs

# Derive
Time (mean ± σ):      3.720 s ±  0.212 s    [User: 9.807 s, System: 1.430 s]
Range (min … max):    3.564 s …  4.298 s    10 runs
```

Every impl is already used by existing tests added in #309, removing any one of them causes those to fail compiling.

Regarding the slice vs array implementation, I looked into it a bit more. From some rudimentary inspection of generated assembly it looks better to use the latter. The `atomic_fence()` call I was so worried about does not actually generate any code at all apparently, and the slice implementation generated a panic branch among other things.

oconnor663:
Oh wow yeah, that's a huge difference. Looks great, thanks again!

Issue #324, opened 2023-07-13
Continual recompliation of blake3


justinabrahms:
Blake3 is in my transitive dependency chain. When I edit my code, it recompiles every time. It seems to believe that the environment variable is changing.. but I'm just editing a file, nothing more.

Most seemingly relevant line below.
```
[2023-07-12T05:41:37Z INFO  cargo::core::compiler::fingerprint]     dirty: EnvVarChanged { name: "CC", old_value: None, new_value: Some("gcc") }
```

Thoughts?

```
$ CARGO_LOG=cargo::core::compiler::fingerprint=info cargo build
[2023-07-12T05:41:37Z INFO  cargo::core::compiler::fingerprint] dependency on `build_script_build` is newer than we are 1689140428.554280381s > 1689140320.333216314s "/home/abrahms/.cargo/registry/src/index.crates.io-6f17d22bba15001f/blake3-1.4.0"
[2023-07-12T05:41:37Z INFO  cargo::core::compiler::fingerprint] fingerprint dirty for WORK_CODE v0.1.0 (/home/abrahms/src/github.com/subconsciousnetwork/WORK_CODE)/Build/TargetInner { name: "WORK_PACKAGE", doc: true, ..: with_path("/home/abrahms/src/github.com/subconsciousnetwork/WORK_CODE/src/main.rs", Edition2021) }
[2023-07-12T05:41:37Z INFO  cargo::core::compiler::fingerprint]     dirty: UnitDependencyInfoChanged { old_name: "cid", old_fingerprint: 2196727044934558216, new_name: "cid", new_fingerprint: 7776841732667268909 }
[2023-07-12T05:41:37Z INFO  cargo::core::compiler::fingerprint] fingerprint dirty for cid v0.10.1/Build/TargetInner { ..: lib_target("cid", ["lib"], "/home/abrahms/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cid-0.10.1/src/lib.rs", Edition2021) }
[2023-07-12T05:41:37Z INFO  cargo::core::compiler::fingerprint]     dirty: UnitDependencyInfoChanged { old_name: "multihash", old_fingerprint: 5789376448465686996, new_name: "multihash", new_fingerprint: 15717749769461962212 }
[...]

PR #413, opened 2024-07-11
Do not let cc-rs trigger a rebuild any time the env changes

hintron:


hintron:
My main goal with this PR is to fix the annoying re-compile/re-check times for Bevy 0.14 on Windows. Bevy 0.14 uses BLAKE3 under the hood (since 0.12, I believe).

`cc-rs` (which BLAKE3 uses) I think [started getting more aggressive at detecting env changes](https://github.com/rust-lang/cc-rs/pull/1103) in v1.1.0 to the point where even `PATH` changes caused cc-rs to want to rebuild, and so that's probably [why I started noticing the issue](https://github.com/bevyengine/bevy/issues/14271) after upgrading to Bevy 0.14.

Luckily, there is a setting to turn off env change rebuilds in `cc-rs`. This should leave it up to BLAKE3 to determine in `build.rs` what envs are important enough to trigger a rebuild for.

For more analysis of what is going on and what I found when digging through the code, see https://github.com/BLAKE3-team/BLAKE3/issues/324#issuecomment-2223573045.

hintron:
Testing this PR results in just the following `rerun-if-env-changed` commands being emitted for me on x64 WIndows 11:

```
cargo:rerun-if-env-changed=CARGO_FEATURE_PURE
cargo:rerun-if-env-changed=CARGO_FEATURE_NO_NEON
cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
cargo:rerun-if-env-changed=CFLAGS_x86_64-pc-windows-msvc
cargo:rerun-if-env-changed=CFLAGS_x86_64_pc_windows_msvc
cargo:rerun-if-env-changed=HOST_CFLAGS
cargo:rerun-if-env-changed=CFLAGS
cargo:rerun-if-env-changed=CARGO_FEATURE_PREFER_INTRINSICS
cargo:rerun-if-env-changed=CARGO_FEATURE_PURE
cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
cargo:rerun-if-env-changed=CARGO_FEATURE_PURE
cargo:rerun-if-env-changed=CARGO_FEATURE_PREFER_INTRINSICS
cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
cargo:rerun-if-env-changed=CARGO_FEATURE_NEON
cargo:rerun-if-env-changed=CARGO_FEATURE_NO_NEON
cargo:rerun-if-env-changed=CARGO_FEATURE_PURE
cargo:rerun-if-env-changed=CC
cargo:rerun-if-env-changed=CFLAGS
```

(the duplicate emissions are a benign issue, but maybe we can clean that up later)

That contains 4 of the 5 envs explicitly checked for in BLAKE's `build.rs`:
* `CARGO_FEATURE_PURE`
* `CARGO_FEATURE_PREFER_INTRINSICS`
* `CARGO_FEATURE_NEON`
* `CARGO_FEATURE_NO_NEON`

All the other env check commands (`CC`, `CFLAGS`, etc.) are still coming from cc-rs, I think, but that's a much better situation than before, where PATH and LIB and all sorts of envs were emitted.

Now, when I make a change to my Bevy game in VSCode with Rust Analyzer (causing `cargo check` to run in the background), I can subsequently run `cargo check` in my terminal without triggering a recompile!

PR #415, opened 2024-07-15
Update the license field in b3sum/Cargo.toml

musicinmybrain:
The top-level `Cargo.toml` was updated to include the `OR Apache-2.0 WITH LLVM-exception` option in b5c6017ea74d7a8be302aefd710dd801c91088a2, but the published crate still just has `CC0-1.0 OR Apache-2.0`. This PR would fix that.


oconnor663:
Thanks!

PR #418, opened 2024-08-18
initial SIMD acceleration for the XOF (AVX-512-only, Unix-only)

oconnor663:
Most of this code has been around for a while on one branch or another, and this PR is a review before I push it to master. All of the assembly was written by @sneves.

### DIFF
--- a/c/blake3_dispatch.c
+++ b/c/blake3_dispatch.c
@@ -234,7 +234,7 @@ void blake3_xof_many(const uint32_t cv[8],
   }
 #if defined(IS_X86)
   const enum cpu_feature features = get_cpu_features();
-#if defined(__unix__) && !defined(BLAKE3_NO_AVX512)
+#if !defined(_WIN32) && !defined(BLAKE3_NO_AVX512)
   if (features & AVX512VL) {
     blake3_xof_many_avx512(cv, block, block_len, counter, flags, out, outblocks);
     return;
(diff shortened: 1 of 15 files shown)


c/blake3_dispatch.c
```diff
@@ -223,6 +223,23 @@ void blake3_compress_xof(const uint32_t cv[8],
...
 }
 
+
+void blake3_xof_many(const uint32_t cv[8],
+                     const uint8_t block[BLAKE3_BLOCK_LEN],
+                     uint8_t block_len, uint64_t counter, uint8_t flags,
+                     uint8_t out[64], size_t outblocks) {
+#if defined(IS_X86)
+  const enum cpu_feature features = get_cpu_features();
+#if defined(__unix__) && !defined(BLAKE3_NO_AVX512)
```
oconnor663, line 233: @sneves what do you think about using `__unix__` here?

sneves: Considering that what is effectively being excluded here is Windows, maybe `!defined(_WIN32)` would make more sense. I believe Apple toolchains may not respect `__unix__` and will define `__APPLE__` or  `__MACH__` or something like that instead.

oconnor663: Done. 4cc34b0e85002f6fef3cd8b264343172cfbc9f58

PR #419, opened 2024-08-21
Add `ConstantTimeEq` support for `Hash`

AaronFeickert:
This PR improves the handling of constant-time equality for hashes.

It adds an optional `subtle` feature that implements `ConstantTimeEq` for `Hash`. This trait uses `subtle` functionality under the hood for the trait implementation.

This also opens the door for other useful `subtle` traits elsewhere in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,16 +41,16 @@ jobs:
       run: cargo run --quiet
       working-directory: ./tools/instruction_set_support
     # Default tests plus Rayon and trait implementations.
-    - run: cargo test --features=rayon,traits-preview,serde,zeroize
+    - run: cargo test --features=rayon,traits-preview,serde,subtle,zeroize
     # Same but with only one thread in the Rayon pool. This can find deadlocks.
     - name: "again with RAYON_NUM_THREADS=1"
-      run: cargo test --features=rayon,traits-preview,serde,zeroize
+      run: cargo test --features=rayon,traits-preview,serde,subtle,zeroize
       env:
         RAYON_NUM_THREADS: 1
     # The mmap feature by itself (update_mmap_rayon is omitted).
     - run: cargo test --features=mmap
     # All public features put together.
-    - run: cargo test --features=mmap,rayon,traits-preview,serde,zeroize
+    - run: cargo test --features=mmap,rayon,traits-preview,serde,subtle,zeroize
     # no_std tests.
     - run: cargo test --no-default-features
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,12 +24,12 @@ neon = []
 # This may become the default in the future.
 wasm32_simd = []
 
-# Enables std::io traits, including impl Write for Hasher, impl Read and Seek
-# for OutputReader, and the Hasher::update_reader method. This feature is
-# enabled by default. (Previously this also controlled runtime CPU feature
-# detection on x86, but now we use the no-std-compatible `cpufeatures` crate
-# for that.)
-std = []
+# This crate uses libstd for std::io trait implementations, and also for
+# runtime CPU feature detection. This feature is enabled by default. If you use
+# --no-default-features, the only way to use the SIMD implementations in this
+# crate is to enable the corresponding instruction sets statically for the
+# entire build, with e.g. RUSTFLAGS="-C target-cpu=native".
+std = ["subtle?/std"]
 
 # The `rayon` feature (disabled by default, but enabled for docs.rs) adds the
 # `update_rayon` and (in combination with `mmap` below) `update_mmap_rayon`
@@ -47,6 +47,9 @@ mmap = ["std", "dep:memmap2"]
 
 # Implement the zeroize::Zeroize trait for types in this crate.
 zeroize = ["dep:zeroize", "arrayvec/zeroize"]
+
+# Enable the `ConstantTimeEq` trait for `Hash` using `subtle` functionality.
+subtle = ["dep:subtle"]
 
 # This crate implements traits from the RustCrypto project, exposed here as the
 # "traits-preview" feature. However, these traits aren't stable, and they're
@@ -118,6 +121,7 @@ digest = { version = "0.10.1", features = ["mac"], optional = true }
 memmap2 = { version = "0.9", optional = true }
 rayon-core = { version = "1.12.1", optional = true }
 serde = { version = "1.0", default-features = false, features = ["derive"], optional = true }
+subtle = { version = "2.6", default-features = false, optional = true }
 zeroize = { version = "1", default-features = false, optional = true }
 
 [target.'cfg(any(target_arch = "x86", target_arch = "x86_64"))'.dependencies]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -143,6 +143,8 @@ use arrayvec::{ArrayString, ArrayVec};
 use core::cmp;
 use core::fmt;
 use platform::{Platform, MAX_SIMD_DEGREE, MAX_SIMD_DEGREE_OR_2};
+#[cfg(feature = "subtle")]
+use subtle::ConstantTimeEq;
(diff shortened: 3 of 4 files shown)


oconnor663:
This change looks good in code, but it seems to come with a substantial performance penalty. I'm pasting this new benchmark at the end of `bench.rs`. (I don't know if all the no_inline/black_box shenanigans are really necessary, but it's a habit.)

```rust
#[no_mangle]
#[inline(never)]
pub fn foo(hash1: &blake3::Hash, hash2: &blake3::Hash) -> bool {
    hash1 == hash2
}

#[bench]
fn bench_eq(b: &mut Bencher) {
    let hash1 = blake3::hash(b"foo");
    let hash2 = blake3::hash(b"bar");
    b.iter(|| foo(test::black_box(&hash1), test::black_box(&hash2)));
}
```

On master I see:

```
$ cargo +nightly bench bench_eq
test bench_eq                          ... bench:           1.37 ns/iter (+/- 0.06)
```

On this branch I see:

```
$ cargo +nightly bench bench_eq
test bench_eq                          ... bench:          31.80 ns/iter (+/- 0.71)
```

31 nanoseconds might not sound like a lot, but for comparison that's almost as long as it takes to _hash_ 64 bytes on my laptop (with AVX-512 support):

```
$ cargo +nightly bench bench_atonce_0001_block
test bench_atonce_0001_block           ... bench:          45.05 ns/iter (+/- 0.33) = 1422 MB/s
```

If we look at `cargo +nightly asm --bench bench foo` and (on your branch) `cargo asm --lib ct_eq` we can see the difference. `constant_time_eq` is able to use SIMD (only SSE2, so there's still room for improvement if we want to get fancy), but `subtle` is doing approximately this for each pair of _bytes_:

```
        movzx eax, byte ptr [r14 + 1]
        xor edi, edi
        cmp al, byte ptr [rbx + 1]
        sete dil
        call subtle::black_box
        mov r15d, eax
        and r15b, bpl
```

Have you looked at these sorts of details before? Do you know if there's anything we can do differently in how we're calling `subtle`, or would this need to be fixed upstream?

> A side benefit is that because subtle plays nicely with Miri, there's no need to keep around the fallback constant-time equality functionality.

This does sound nice. If we don't end up landing this change, I might try to contribute something similar to `constant_time_eq`.

AaronFeickert:
The optimization barrier in `subtle` used to be inline assembly, but it was changed a few years back (and has seen some iteration). I've seen discussion about returning to that behavior, with initial concerns about stability and portability. Unfortunately, the use of `subtle` would currently require the performance hit. That being said, it is pretty standard at this point in the Rust cryptography ecosystem.

The core library does include its own [optimization barrier](https://doc.rust-lang.org/stable/core/hint/fn.black_box.html) that was previously available in `subtle`. However, a warning was added not to use it for cryptographic purposes, so it's no longer made available.

One option that I mulled over was whether it's reasonably safe enough to use variable-time equality testing for `PartialEq`, and reserve constant-time equality testing for `ConstantTimeEq`. This has the downside of being breaking and easier to misuse, but it does allow implementers to decide if their use case requires it for improved flexibility.

AaronFeickert:
Anyway, I'm happy to close this if the performance hit isn't worth the library change and Miri cleanup.

oconnor663:
Let's see what Cesar thinks of that PR I just put up. Also if some callers need trait compatibility with `subtle::ConstantTimeEq`, another option could be to make `subtle` an optional feature and implement it as a wrapper around the current implementation.

AaronFeickert:
Is there a particular reason why the current Miri workaround doesn't just use variable-time equality, if it's not intended for use outside of testing (and not used for benchmarks)? Having a pseudo-constant-time variant in the codebase for that purpose seems like a small footgun that someone might use elsewhere unsafely.

oconnor663:
No particularly good reason, you're right. It might make sense to delete it, but on the other hand there's some copy-paste risk on both sides of the equation. We don't want folks to use the naive function thinking it's constant time, but we also don't want folks to think it's ok to use non-constant-time comparisons with secret values. Either way we kind of have to lean on scary comments.

AaronFeickert:
Good point overall; the only proper approach would be to nix it entirely, so hopefully your PR gets accepted!

I do think that the overall risk is still greater (at least marginally) with the current workaround, especially given that the function is named [`constant_time_eq_miri`](https://github.com/BLAKE3-team/BLAKE3/blob/479eef82d756221abe9f93077b91ce8cf763c32a/src/lib.rs#L322) and there's no reason to think it couldn't or wouldn't ever be variable time. Unless it's too much of a nit, I'll go ahead and open a PR to remove it now in favor of standard variable-time `PartialEq` on the underlying byte array; a scary comment will accompany it!

Separately, good idea that `subtle` could be used as an optional feature, to get its useful traits for callers who need or want them. If that were the case, I think it would be proper to have its constant-time trait implementations use `subtle` directly and _not_ wrap `constant_time_eq`, in order to ensure the caller is getting what it says on the box. This would have the extra advantage of allowing the existing `PartialEq` functionality to retain its speedy behavior. The extra cost would be the extra dependency, but at least it would be optional.

cesarb:
> Let's see what Cesar thinks of that PR I just put up.

Sorry for the delay. Looks fine to me. Merged and published `constant_time_eq` 0.3.1 with only that change.

AaronFeickert:
The change that @cesarb references is now accounted for in #421.

oconnor663:
Thanks @AaronFeickert and @cesarb. Merged!

AaronFeickert:
@oconnor663: After the ongoing discussion, this has been updated to retain the existing (speedy) functionality from `constant_time_eq`, but add a `ConstantTimeEq` implementation for `Hash` via a new optional `subtle` trait.

I think this is a useful addition, and also opens the door to adding other useful traits in the future if desired.

AaronFeickert:
Can a maintainer re-run the failed tests?

BurningEnlightenment:
@AaronFeickert there you go. Feel free to ping me if you need something else 🙂

AaronFeickert:
@BurningEnlightenment: Thanks! This PR has been in the queue for the while; would appreciate any insight into whether or not it is useful enough to keep around.

BurningEnlightenment:
I am in favor of merging, but the decision is not up to me.

oconnor663:
Sorry for letting this sit. Two questions: Even though we're implementing the trait from `subtle`, can we use `constant_time_eq` to do the comparison internally? And should we fold this into the existing `traits-preview` feature, which enables trait implementations from the `digest` crate? I know that `digest` is a RustCrypto project and `subtle` is not, but on the other hand `digest` does take a dependency on `subtle`.

AaronFeickert:
We certainly could use the existing `constant_time_eq` functionality, but that doesn't really sit well with me. There's no strict requirement that `ConstantTimeEq` be implemented in any particular way (aside from attempting to execute in constant time), but I wager users of the library would expect that a `subtle` feature implementing `subtle` traits would use `subtle` functionality where available, especially since their use of those traits likely implies their trust in `subtle` functionality.

I also don't think baking it in with the existing `traits-preview` feature is the right move, since `subtle` is stable and independent from the digest traits. Keeping it separate also has the advantage of allowing for a (slightly) smaller dependency footprint.

But I am open to discussion on these points!

oconnor663:
For more context for me, what's an example of a caller that needs an implementation of the `Subtle` trait specifically (i.e. in a generic context), as opposed to just "we promise equality is constant time" in the docs?

AaronFeickert:
I don't think any caller _needs_ the `subtle`-specific implementation of constant-time operations. But it seems somewhat off to provide `subtle` constant-time traits but not the corresponding implementations. Are you instead asking why we should bother providing `ConstantTimeEq` as a trait at all?

Aside from this particular PR, I'm also interested to propose constant-time ordering that uses `subtle` traits. Doing so using `subtle` functionality will be much cleaner using its implementations as well, and maintaining consistency on this front seems prudent to me.

PR #420, opened 2024-08-21
Implement constant-time hash ordering

AaronFeickert:
This PR implements constant-time ordering (and partial ordering) for `Hash`, but does so differently than #267 and #370.

It uses [`subtle`](https://crates.io/crates/subtle) internally to [lexicographically](https://doc.rust-lang.org/std/cmp/trait.Ord.html#lexicographical-comparison) compare corresponding bytes, being careful not to short circuit. This simplifies the logic considerably.

Note that it does not use `subtle` to implement `PartialEq`; this functionality exists in #419 since it seemed better scoped for a separate PR. However, if #419 is implemented, it will be possible to implement `ConstantTimeGreater` and `ConstantTimeLess` due to the latter requiring a `ConstantTimeEq` trait bound included in that PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ neon = []
 # --no-default-features, the only way to use the SIMD implementations in this
 # crate is to enable the corresponding instruction sets statically for the
 # entire build, with e.g. RUSTFLAGS="-C target-cpu=native".
-std = []
+std = ["subtle/std"]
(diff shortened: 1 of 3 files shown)


AaronFeickert:
Marking as draft for now, since #419 would add `subtle` support via an optional feature that would presumably be used here.

PR #421, opened 2024-08-22
Remove Miri equality workaround function

AaronFeickert:
When testing with Miri, hash equality is checked using a [workaround function](https://github.com/BLAKE3-team/BLAKE3/blob/479eef82d756221abe9f93077b91ce8cf763c32a/src/lib.rs#L318-L331) that attempts to mimic constant-time functionality. This was done only because the [`constant_time_eq`](https://crates.io/crates/constant_time_eq) crate was not compatible with Miri.

However, thanks to a [PR](https://github.com/cesarb/constant_time_eq/pull/11) by @oconnor663, this incompatibility is no longer present, which means the workaround can be safely removed. This PR does so.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,7 +97,7 @@ features = ["mmap", "rayon", "serde", "zeroize"]
 [dependencies]
 arrayref = "0.3.5"
 arrayvec = { version = "0.7.4", default-features = false }
-constant_time_eq = "0.3.0"
+constant_time_eq = { version = "0.3.1", default-features = false }
 cfg-if = "1.0.0"
 digest = { version = "0.10.1", features = [ "mac" ], optional = true }
 memmap2 = { version = "0.9", optional = true }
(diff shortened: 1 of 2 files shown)


oconnor663:
LGTM, though it looks like it needs a tweak to pass CI:

```
error[E0277]: can't compare `[u8; 32]` with `&[u8; 32]`
   --> src/lib.rs:339:23
    |
339 |         return self.0 == other;
    |                       ^^ no implementation for `[u8; 32] == &[u8; 32]`
    |
    = help: the trait `std::cmp::PartialEq<&[u8; 32]>` is not implemented for `[u8; 32]`
help: consider dereferencing here
    |
339 |         return self.0 == *other;
    |                          +
```

AaronFeickert:
Good catch! I'm surprised that my IDE didn't complain about this. Will update.

AaronFeickert:
Should be good to go now.

AaronFeickert:
@oconnor663: This is now updated to account for the new `constant_time_eq` version that includes your fix.

oconnor663:
Thanks!

Issue #13, opened 2020-01-10
Hasher::new_derive_key accepts a str, why not [u8]?


WildCryptoFox:
There is no reason for UTF-8 here. Please `s/str/[u8]/` here. Alternatively provide another function to avoid the breaking change. Though, as I've mentioned in #11 I'm not against a breaking change for consistency with other crates, especially this early.

https://docs.rs/blake3/0.1.0/blake3/struct.Hasher.html#method.new_derive_key

```rust
pub fn new_derive_key(context: &str) -> Self
```

Luro02:
You could also change the signature to
```rust
new_derive_key<T: AsRef<[u8]>>(context: T) -> Self
```
which would allow vec and slices

WildCryptoFox:
@Luro02 That's also almost a non-breaking change.. but there are cases which would break (I.e. `Hasher::new_derive_key(foo.into())`. Could also use `(context: impl AsRef<[u8]>)`. I usually do this, surprised I didn't suggest it in the top post. :)

oconnor663:
`&str` (and `char*` in the C code) was a deliberate choice in this implementation (not required by the spec) to discourage misuse. For all the details, please take a look at Sections 6.2 and 7.8 in [the spec](https://github.com/BLAKE3-team/BLAKE3-specs/blob/master/blake3.pdf) and at the [summarized version](https://docs.rs/blake3/0.1.0/blake3/fn.derive_key.html) in the Rust docs. I should be upfront that this is something that we knew in advance a lot of programmers would find weird, and that a lot of people would think the problem we're trying to solve isn't a big problem. So with that in mind, I'll describe what we mean by "misuse". Forgive me for what's going to be a long comment.

The way `derive_key` is supposed to be used is with a constant context string, like this:

```rust
const XYZ_CONTEXT_STRING: &str = "foo bar app 2020-01-10 14:54:14 purpose xyz";

fn purpose_xyz_subkey(root_key: &[u8; 32]) -> [u8; 32] {
    let mut out = [0; 32];
    blake3::derive_key(XYZ_CONTEXT_STRING, root_key, &mut out);
    out
}
```

Of course not everyone is going to follow that exact format for the context string, I'm not kidding myself, but the main thing is that I've picked something that no one else will pick and that I've hardcoded it. In particular, `derive_key` is **not** supposed to be used with a dynamic context string, like this:

```rust
[...]

PR #424, opened 2024-08-28
derive_key: Add default-off, unstable `derive_key_u8ref` API call

jac-cbi:
This can be enabled with the unstable feature flag, `i_know_what_i_am_doing`.

Developers attempting to deploy blake3 via the Rust crate often encounter `derive_key(&str, ...)` which forces the INFO parameter to be a valid UTF-8 string.

Please see the discussion in issue [13](https://github.com/BLAKE3-team/BLAKE3/issues/13), in particular, this [comment](https://github.com/BLAKE3-team/BLAKE3/issues/13#issuecomment-573219903).

The recommended course of action for those with non-UTF-8 INFO is to use `hash` and `keyed_hash` and open code their own `derive_key` which takes an `&[u8]`.

This is not good for two reasons:  First, it is quickly seen that this forces deviation from the Blake3 paper for how `derive_key` should be performed, as `hash` doesn't let you set the `flags` field.  Attempting to use the underlying `hash_all_at_once` fails because it's not exported.  Second, the developer is now forced into the position of maintaining their own patches on top of the blake3 repo.  This is a burden on the developer, and makes following blake3 upstream releases *much* less likely.

This patch proposes a reasonable compromise.  For developers who require `&[u8]` INFO field, they can enable the rust feature flag `i_know_what_i_am_doing` to expose the API `derive_key_u8ref`.  This enables developers to use upstream blake3 directly, while still discouraging sloppy behavior in the default case.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,6 +90,17 @@ no_avx2 = []
 no_avx512 = []
 no_neon = []
 
+# For developers who encountered `derive_key(&str)`, read the Blake3 paper,
+# and read and fully comprehended this comment:
+#
+#   https://github.com/BLAKE3-team/BLAKE3/issues/13#issuecomment-573219903
+#
+# and are willing to own both halves when their design breaks.
+#
+# Note: The *only* reason to add this is so that developers don't need to
+# maintain out-of-tree hacked up patches.
+i_know_what_i_am_doing = []
+
 [package.metadata.docs.rs]
 # Document the rayon/mmap methods and the Serialize/Deserialize/Zeroize impls on docs.rs.
 features = ["mmap", "rayon", "serde", "zeroize"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -990,6 +990,16 @@ pub fn derive_key(context: &str, key_material: &[u8]) -> [u8; OUT_LEN] {
         .0
 }
 
+#[cfg(i_know_what_im_doing)]
+pub fn derive_key_u8ref(context: &[u8], key_material: &[u8]) -> [u8; OUT_LEN] {
+    let context_key =
+        hash_all_at_once::<join::SerialJoin>(context, IV, DERIVE_KEY_CONTEXT).root_hash();
+    let context_key_words = platform::words_from_le_bytes_32(context_key.as_bytes());
+    hash_all_at_once::<join::SerialJoin>(key_material, &context_key_words, DERIVE_KEY_MATERIAL)
+        .root_hash()
+        .0
+}
+
 fn parent_node_output(
     left_child: &CVBytes,
     right_child: &CVBytes,


oconnor663:
> workflow awaiting approval

You didn't do anything wrong, that's just a change GitHub made a couple years back where they stopped automatically running CI on PRs from new contributors. Apparently some people were "contributing" cryptocurrency mining code to random repos and abusing GitHub compute. Something like that. Anyway, I've clicked the run button :)

oconnor663:
> The recommended course of action for those with non-UTF-8 INFO is to use `hash` and `keyed_hash` and open code their own `derive_key` which takes an `&[u8]`.

I agree that this is a thing you can do, and even a thing that I suggested in that thread :) but I might not go quite as far as "recommended course of action" without hearing more about what a specific caller is trying to accomplish.

> `hash` doesn't let you set the `flags` field

Indeed, nothing lets you set that field, and I'd be surprised if we ever exposed a public API for it. To kind of repeat and summarize my wall of text from #13, the goal of `derive_key` is to help _uncoordinated_ callers who happen to share key material from accidentally having bad interactions (reusing each other's nonces with the same key, publishing each other's intermediate secrets etc). Of course if two callers can coordinate with each other, then domain separation is easy: you prefix all your inputs with "foo", and I'll prefix all my inputs with "bar", and boom we're domain separated. You don't need any special modes or features for that. It's when there's no coordination that you have to rely on requirements and conventions to keep you safe. And so the value of `derive_key` is tied to its requirements and conventions, that plus the fact that we internally domain-separate it from the other APIs that don't have the same requirements.

> For developers who require `&[u8]` INFO field

My suspicion for a long time is that needing arbitrary bytes is usually a side effect of violating the "hardcoded" requirement of the `derive_key` context string. Could you tell me more about your use case?

I should mention that we do provide a C API that takes bytes, `blake3_hasher_init_derive_key_raw`. The issue there is that ordinary ASCII/UTF-8 C strings are null-terminated, but strings in most other languages are not, and we didn't want to force bindings to heap allocate and copy strings just to add the null terminator. However, Rust doesn't have the same problem. As you probably already know, `std::str::from_utf8` doesn't require a null terminator and doesn't do any heap allocation, and you can cheaply convert a `&[u8]` to a `&str`. If you're completely sure the bytes are valid UTF-8, you can even do a zero-cost `unsafe` conversion. But if you're not sure...I'd like to hear more about why you're not sure.

One scenario I could imagine (not suggesting this is your scenario, just making stuff up here) is using BLAKE3 to implement API-compatible replacement for an existing HKDF construction that currently takes its info/ikm/salt parameters as raw bytes. I'd suggest that `derive_key` is _not_ a good fit for building such a replacement, precisely because it imposes requirements on the context string that the previous API did not impose.

oconnor663:
Oops, thanks for the ping.

> in order to incorporate context-distinguishing, non-private data, I'm looking to concatenate some unique info about the environment of the context establishment. Things like public IP / port of both peers with epoch time in seconds.

Gotcha. I think what I'd recommend here is a two-step process along the lines of: 1) `derive_key` your long-lived context data as _input bytes_, using a hardcoded context string, producing an intermediate 32-byte key. Then 2) `keyed_hash` your shorter-lived inputs with that intermediate key. If your design doesn't need more than one context string for step 1, you can also replace `derive_key` with just `hash`. That `hash`+`keyed_hash` construction would be [identical to what `derive_key` does internally](https://github.com/BLAKE3-team/BLAKE3/blob/60ff2eafed63b29ed1622bb6330e640c22c683ff/c/blake3.c#L387-L394). The only differences are that it doesn't use the internal domain-separation bitflags that are unique to `derive_key`, and that (for that reason) it's not subject to the security requirement that you must hardcode the first argument.

Thinking about your example is helping me clarify in my head what `derive_key` is trying to do. It _looks_ like it's solving two different problems at once:

1. Deriving independent subkeys from a shared secret, to help you use the same secret safely with different algorithms, or with different instances of the same algorithm that shouldn't interact for whatever reason.
2. Hashing two arbitrary-length strings together unambiguously, especially-but-maybe-not-exclusively when one of those strings is some sort of key.

But in fact, `derive_key` is only intended for the first thing, not the second thing. There's nothing _wrong_ with the second thing per se, but the hardcoded context requirement is important for the first thing, and unfortunately it's incompatible with the second thing.

I think all of this provokes a very natural follow up question, and if you'll forgive me, I'll express that question as a meme:

[image](https://twitter.com/jrecursive/status/1414067934381547527)

Am I really telling folks to hand-roll their own [`TupleHash`](https://www.nist.gov/publications/sha-3-derived-functions-cshake-kmac-tuplehash-and-parallelhash)? Well...yes...I am. Which is lame. And I'm ashamed. But in my defense, even if we said `derive_key` was a fine way to hash any two strings, we'd still be telling folks to figure things out for themselves as soon as they had _three_ strings. Or consider another flavor of the same question: Do the hashes of these two structs need to be domain-separated from each other?

```
struct User {
    name: String,
    age: u32,
}

struct Server {
    domain: String,
    port: u32,
}
```

`TupleHash` doesn't know. It _can't_ know. Different applications need different things. If we came up with a fully generic solution (`#[derive(HashMyStruct)]`) that incorporated every possible bit of local context, then many users would be dismayed to discover that they can no longer change the _names_ of their structs or fields without breaking backwards compatibility. It's surprisingly difficult to come up with a general solution to this, much like it's surprisingly difficult to come up with a one-size-fits-all serialization format.

Another related question, how do we compare these three different approaches to the same problem:

1. `keyed_hash(key=hash(A), input=B)`
2. `hash(hash(A) + hash(B))`
3. `hash(A + B + len(B))`

All of these are valid ways to unambiguously hash `A` and `B` together, with a hash like BLAKE3 or SHA-3 that doesn't allow length extension. I recommended the first approach because it's slightly more efficient when `A` rarely changes. An advantage of the second approach is that, if `A` and `B` are both very long, it's possible to start processing `B` before you've gotten to the end of `A`, or to process `A` and `B` on different threads. The third approach wins when `A` and `B` are both very short, because it's the only option that can reduce to a single call to the compression function. I think there's an argument to be made that some hybrid approach might be best, which dynamically switches between the third and the second as inputs grow. But for any given application, it's probably best to keep it simple and pick one of those three based on whatever feels right for the problem at hand?

PR #426, opened 2024-09-06
build(CMake): Fix pkg-config for absolute CMAKE_INSTALL_*DIR

fgaz:
CMAKE_INSTALL_*DIR can be absolute, and in that case ${prefix} should not be prepended.

See https://github.com/jtojnar/cmake-snips/?tab=readme-ov-file#concatenating-paths-when-building-pkg-config-files

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -229,6 +229,31 @@ install(FILES
     DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/blake3"
 )
 
+# Function for joining paths known from most languages
+#
+# SPDX-License-Identifier: (MIT OR CC0-1.0)
+# Copyright 2020 Jan Tojnar
+# https://github.com/jtojnar/cmake-snips
+#
+# Modelled after Python’s os.path.join
+# https://docs.python.org/3.7/library/os.path.html#os.path.join
+# Windows not supported
+function(join_paths joined_path first_path_segment)
+    set(temp_path "${first_path_segment}")
+    foreach(current_segment IN LISTS ARGN)
+        if(NOT ("${current_segment}" STREQUAL ""))
+            if(IS_ABSOLUTE "${current_segment}")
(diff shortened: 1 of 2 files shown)


c/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
-cmake_minimum_required(VERSION 3.9 FATAL_ERROR)
+# For cmake_path function
+cmake_minimum_required(VERSION 3.20 FATAL_ERROR)
```
BurningEnlightenment, line 2: RHEL 7 is still in extended support. The next upgrade would be `3.11` as that's what RHEL 8 ships with.
So unless the current CMake version deprecates compatibility with v3.9 (as with v3.5), I'm unwilling to raise this requirement.

BurningEnlightenment (changes requested):

c/CMakeLists.txt
fgaz: I can use [this module](https://github.com/jtojnar/cmake-snips/blob/master/CMakeScripts/JoinPaths.cmake) instead. Is that ok?

BurningEnlightenment: Sure, but please embed the function (and the copyright notice, the license is compatible) in the `CMakeLists.txt`.
Furthermore I'd request the new variables to be written in all-caps and maybe renamed to `PKG_CONFIG_INSTALL_*` mirroring the `GNUInstallDirs` naming scheme.

fgaz: Done

BurningEnlightenment (approved):

fgaz:
Is there anything I can do to move this forward?

BurningEnlightenment:
If @oconnor663 doesn't say no until Friday, I'll go ahead and merge this. I don't think that his review is strictly necessary in this case.

fgaz:
friendly ping

BurningEnlightenment:
Thanks!

PR #428, opened 2024-09-23
support win7 target

zhangli-pear:
Add support for [*-win7-windows-msvc](https://doc.rust-lang.org/rustc/platform-support/win7-windows-msvc.html#-win7-windows-msvc).

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -80,7 +80,7 @@ fn is_big_endian() -> bool {
 // disabled by old compilers.)
 fn is_windows_msvc() -> bool {
     // Some targets are only two components long, so check in steps.
-    target_components()[1] == "pc"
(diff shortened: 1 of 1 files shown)


Fenex:
I've tried to compile a modern rustc as host compiler for running on win7 and facing with the same error.

However I think its will be better to skip checks for `target_components()[1]` (known as `target_vendor` at all) due vendors can be various, but this all is msvc actually. Before we have "pc" only, right now we want to add "win7". Perhaps, in the future we need to add something else. At least for now there is exists a "[rust9x](https://github.com/rust9x/rust/commit/3df0d56049f2411bde914dcb531c822e563f5266)" target vendor to compile for Windows Vista and older (down to 9x/ME).

PR #439, opened 2025-01-05
Make portable implementation `const`

### LINKED ISSUE #440: BLAKE3 in `const` environment
I was going to use BLAKE3 in `const` environment, but it doesn't support such use case yet.

https://github.com/BLAKE3-team/BLAKE3/pull/439 make underlying portable implementation compatible with `const`, but didn't expose it publicly.

I wasn't sure what API you'd like and whether it should be feature-gated or not. Something like `blake3::hash_const()` or `blake3::const_hash()` might be reasonable, making `portable` module public and adding a separate `blake3::portable::hash()` (which would implicitly support `const`) there would make sense too.

The main complication is that Rust doesn't compile the library separately for `const`, so we can't use the generic code path that uses inline assembly and other features that const environment doesn't support yet, it has to be a separate entrypoint.

Looking for feedback and happy to open follow-up PR with the desired changes.

nazar-pc:
I was going to use BLAKE3 in `const` environment, but it doesn't support such use case yet.

This relatively small PR refactors portable implementation to make it work in `const` environment.

There is a bit of trivially verifiable `unsafe` due to nicer APIs not yet available in `const` environment, but these all work on stable.

`arrayref` were hard to read and used slicing that is not compatible with `const`, so I replaced it with a couple simple macros where it was necessary and left other usages as is.

---

UPD: I decided to implement the whole thing, added `const_hash`, `const_keyed_hash` and `const_derive_key` functions by copying the originals and modifying them to use `const fn` portable implementation.

I didn't add `ConstHasher`, though it is possible too if someone needs it.

Resolves https://github.com/BLAKE3-team/BLAKE3/issues/440

### DIFF
--- a/src/platform.rs
+++ b/src/platform.rs
@@ -1,5 +1,4 @@
 use crate::{portable, CVWords, IncrementCounter, BLOCK_LEN};
-use arrayref::{array_mut_ref, array_ref};
 
 cfg_if::cfg_if! {
     if #[cfg(any(target_arch = "x86", target_arch = "x86_64"))] {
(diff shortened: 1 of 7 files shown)


src/platform.rs
```diff
@@ -473,74 +472,96 @@ pub fn sse2_detected() -> bool {
...
+            $dst[$chunk_index * 4 + 0],
+            $dst[$chunk_index * 4 + 1],
+            $dst[$chunk_index * 4 + 2],
+            $dst[$chunk_index * 4 + 3],
+        ] = $src[$chunk_index].to_le_bytes();
+    };
+}
+
 #[inline(always)]
 pub fn words_from_le_bytes_32(bytes: &[u8; 32]) -> [u32; 8] {
```
devinrsmith, line 498: Missing `const` here?

src/portable.rs
```diff
@@ -120,7 +119,7 @@ pub fn compress_xof(
     crate::platform::le_bytes_from_words_64(&state)
 }
 
-pub fn hash1<const N: usize>(
+const fn hash1<const N: usize>(
```
devinrsmith, line 122: Removal of `pub` intended?

devinrsmith (commented):
FYI: I'm not a maintainer of this library, but I do like to keep up to date with some of the going-ons.

src/platform.rs
nazar-pc: Wasn't needed, but will add for consistency with other functions

src/portable.rs
nazar-pc: Not really, but since there are a few other backends with the same function and in all cases it is internal implementation without `pub` (here too), I decided to remove it for consistency too. Happy to restore if maintainers have strong opinion on this one.

nazar-pc (commented):
Thanks for review, appreciate it!

nazar-pc:
@oconnor663 it'd be great to get your attention here, especially since the changes are fairly simple

nazar-pc:
Ping :slightly_smiling_face:

nazar-pc:
I have implemented public const functions too and added them to tests against reference implementation, everything passes. Just rebased against `master` to pull Rust 1.78.0 to get some of the features that were previously unstable on older version of Rust.

@oconnor663 would be great if you can find time to look at this, especially since all tests are passing.

nazar-pc:
Hm... I see https://github.com/BLAKE3-team/BLAKE3/commit/a31e519869d5751370f50c39a99340660ee95bf7 says that true MSRV is 1.66.0, if that is the case I can put new features behind feature flag or maybe MSRV can be increased (1.77.0 is what is needed here I believe, which was released almost a year ago).

oconnor663:
Apologies for not reviewing this more promptly. You might already have noticed that I just pushed the CI-tested MSRV to 1.78 (for the `ignore` crate).

nazar-pc:
Just discovered 1.83.0 is what is needed for `ptr::copy_to_nonoverlapping()`

nazar-pc:
Okay, so `&mut` is not allowed in `const fn` until current stable 1.85.0. Sorry for the noise here, I added `const` feature that enabled these new APIs, retaining support for older versions of Rust.

oconnor663:
I'm leaning against taking these changes. It looks like a lot of complicated logic has to be duplicated, and much of that logic (like `compress_subtree_wide`) is only as complicated as it is because it supports SIMD parallelism, which the `const` version gets no benefit from. I wonder if it could make more sense to fork the `reference_impl` rather than the highly optimized implementation?

That said, my gut level, haven't-spent-much-time-with-the-feature impression of const generics is that it's still too early to rely heavily on them. Errors like this feel silly to me:

```
error[E0015]: cannot use `for` loop on `std::ops::Range<usize>` in constant functions
  --> reference_impl.rs:68:14
   |
68 |     for i in 0..16 {
   |              ^^^^^
```

Another minimim-viable-feature for me would be basic arithmetic in const parameters, which doesn't work today:

```rust
struct Hasher<const LEN: usize> {
    state: [u8; LEN * 2],  // error: generic parameters may not be used in const operations
}
```

Does all that sound reasonable?

somethingelseentirely:
Any update on this 😄?
We're computing persistent identifiers in our codebase and it would be really nice to be able to make them `const`.

oconnor663:
I'm still mostly against making this change within this crate, but I think a `const` fork of the portable implementation or the reference implementation makes sense, and I'd be happy to review it and link to it from our README. I think there are enough `const` functions already in place on `blake3::Hash` that the return types could interoperate, but if we need to make tweaks there that could also make sense.

somethingelseentirely:
Yeah I agree that it probably makes the most sense to create a const version of the reference implementation, given that performance is mostly not an issue for `const` use cases afaict. Most things that you want to hash at compile time are probably going to be quite small in size and, aside from compile time convenience, not performance critical anyways.

PR #442, opened 2025-01-08
Add keyed and derive_key variants for guts

devinrsmith:
This adds a struct `Guts` with methods `new`, `new_keyed`, and `new_derived_key` mirroring `Hasher` construction options.

### DIFF
--- a/src/guts.rs
+++ b/src/guts.rs
@@ -9,19 +9,84 @@
 pub const BLOCK_LEN: usize = 64;
 pub const CHUNK_LEN: usize = 1024;
 
+#[derive(Clone, Copy, Debug)]
+pub struct Guts {
+    cv: crate::CVWords,
+    flags: u8,
+    platform: crate::platform::Platform,
+}
+
+impl Guts {
+    fn new_internal(key: &crate::CVWords, flags: u8) -> Self {
+        Self {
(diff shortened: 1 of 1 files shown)


oconnor663:
I've been planning a more extensive change here, where the `guts` module will provide functions that let you initialize or update a regular `Hasher` in advanced ways. That should support fancy use cases while still taking advantage of the full SIMD/threading performance of `Hasher`, without needing to reinvent the wheel.

devinrsmith:
Do you imagine that the guts advanced `Hasher` would be enough to support the bao implementation, or would bao still need something lower-level?

Any estimate on timeline for the more extensive changes?

I don't expect API stability for guts today; and if you are amenable, would be happy for this to be merged even if it will be replaced in the future.

oconnor663:
> Do you imagine that the guts advanced Hasher would be enough to support the bao implementation

Yes, the goal here would be to at least support:

- Bao as it currently works today
- plus the keyed and derive_key modes, which Bao doesn't currently use but could in theory
- plus a "pre-process the derive_key context string" option

> Any estimate on timeline for the more extensive changes?

No I don't have a timeline, but I'll make a calendar note to prioritize it this weekend :)

PR #446, opened 2025-02-12
Implement llfio-based memory-mapped IO for C lib

silvanshade:
This PR is a follow up to #445 and implements [llfio](https://github.com/ned14/llfio)-based memory-mapped IO.

A new example using this functionality has been added as `example-mmap.c`.

The example can be compiled, installed, and run with the following commands:

```bash
# Also add -DBLAKE3_FETCH_TBB=1 if compiling on Windows
$ cmake -S c -B c/build -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=c/build/install -DBLAKE3_USE_LLFIO=1 -DBLAKE3_USE_TBB=1 -DBLAKE3_EXAMPLES=1
$ cmake --build c/build --target install
$ c/build/install/bin/blake3-example-mmap <path>
```

### DIFF
--- a/c/blake3.h
+++ b/c/blake3.h
@@ -27,6 +27,12 @@
 #endif
 
 #ifdef __cplusplus
+#define BLAKE3_NOEXCEPT noexcept
+#else
+#define BLAKE3_NOEXCEPT
+#endif
+
+#ifdef __cplusplus
 extern "C" {
 #endif
 
@@ -69,9 +75,17 @@ BLAKE3_API void blake3_hasher_init_derive_key_raw(blake3_hasher *self, const voi
                                                   size_t context_len);
 BLAKE3_API void blake3_hasher_update(blake3_hasher *self, const void *input,
                                      size_t input_len);
+#if defined(BLAKE3_USE_LLFIO)
+BLAKE3_API int blake3_hasher_update_mmap(blake3_hasher *self,
+                                         const char *path) BLAKE3_NOEXCEPT;
+#endif // BLAKE3_USE_LLFIO
 #if defined(BLAKE3_USE_TBB)
 BLAKE3_API void blake3_hasher_update_tbb(blake3_hasher *self, const void *input,
                                          size_t input_len);
+#if defined(BLAKE3_USE_LLFIO)
+BLAKE3_API int blake3_hasher_update_mmap_tbb(blake3_hasher *self,
+                                             const char *path) BLAKE3_NOEXCEPT;
+#endif // BLAKE3_USE_LLFIO
 #endif // BLAKE3_USE_TBB
 BLAKE3_API void blake3_hasher_finalize(const blake3_hasher *self, uint8_t *out,
                                        size_t out_len);
--- a/c/blake3_llfio.cpp
+++ b/c/blake3_llfio.cpp
@@ -1,0 +1,142 @@
+#include <array>
+#include <utility>
+#include <variant>
+
+#include <llfio.hpp>
+#include <outcome.hpp>
+#include <quickcpplib/signal_guard.hpp>
+
+#include "blake3.h"
+#include "blake3_impl.h"
+
+#ifdef _WIN32
+#include <windows.h>
+#include <winternl.h>
+#endif
+
+namespace llfio = LLFIO_V2_NAMESPACE;
+namespace quickcpp = QUICKCPPLIB_NAMESPACE;
+
+namespace {
+template <class... Ts> struct overloads : Ts... {
+  using Ts::operator()...;
+};
+// NOTE: Deduction guide only needed for Apple Clang.
+template <class... Ts> overloads(Ts...) -> overloads<Ts...>;
+
+template <class T> auto error_code(T &&error) -> long long {
+  const auto value = error.value();
+#if LLFIO_EXPERIMENTAL_STATUS_CODE
+  return value.sc;
+#elif !defined(_WIN32)
+  return value;
+#else
+  return NT_ERROR(value) ? RtlNtStatusToDosError(value) : value;
+#endif
+}
+
+namespace {
+const quickcpp::signal_guard::signalc_set guarded_signals =
+    quickcpp::signal_guard::signalc_set::undefined_memory_access // SIGBUS
+    | quickcpp::signal_guard::signalc_set::segmentation_fault    // SIGSEGV
+    ;
+
+// Install and enable global signal handlers for specified signals.
+static const quickcpp::signal_guard::signal_guard_install
+    signal_guard_install(guarded_signals);
+} // namespace
+} // namespace
+
+INLINE auto copy_wide(blake3_hasher *self, llfio::file_handle &file) noexcept
+    -> llfio::result<void> {
+  std::array<std::byte, 65536> buffer;
+  std::uint64_t offset = 0;
+  while (true) {
+    std::uint64_t bytes_read = 0;
+    if (auto result = llfio::read(file, offset, {buffer})) {
+      bytes_read = result.assume_value();
+      blake3_hasher_update(self, buffer.data(), bytes_read);
+      offset += bytes_read;
+    } else {
+#ifdef _WIN32
+      if (error_code(result.assume_error()) == ERROR_HANDLE_EOF) {
+        break;
+      }
+#endif
+      if (result.assume_error() != llfio::errc::interrupted) {
+        return std::move(result).as_failure();
+      }
+
+      continue;
+    }
+
+    if (bytes_read < buffer.size()) {
+      break;
+    }
+  }
+  return llfio::success();
+}
+
+INLINE auto maybe_mmap_path(char const *path) noexcept -> llfio::result<
+    std::variant<llfio::mapped_file_handle, llfio::file_handle>> {
+  const auto base_dir = llfio::path_handle{};
+  if (auto mmap = llfio::mapped_file(base_dir, path); mmap.has_value()) {
+    return std::move(mmap).value();
+  }
+  // Memory mapping may fail so fall back to normal file reading if necessary.
+  OUTCOME_TRY(auto file, llfio::file(base_dir, path));
+  return file;
+}
+
+INLINE int blake3_hasher_update_mmap_base(blake3_hasher *self, char const *path,
+                                          bool use_tbb) noexcept {
+  auto result = [=]() -> llfio::result<void> {
+    OUTCOME_TRY(auto handle, maybe_mmap_path(path));
+    OUTCOME_TRY(std::visit(
+        overloads{
+            [=](llfio::mapped_file_handle &mmap) -> llfio::result<void> {
+              OUTCOME_TRY(const auto extent, mmap.maximum_extent());
+              const auto result = quickcpp::signal_guard::signal_guard(
+                  guarded_signals,
+                  [&]() -> bool {
+                    const auto data = mmap.address();
+                    blake3_hasher_update_base(self, data, extent, use_tbb);
+                    return true;
+                  },
+                  [=]([[maybe_unused]] const auto *info) -> bool {
+                    return false;
+                  });
+              if (!result) {
+                return llfio::errc::io_error;
+              }
+              return llfio::success();
+            },
+            [=](llfio::file_handle &file) -> llfio::result<void> {
+              OUTCOME_TRY(copy_wide(self, file));
+              return llfio::success();
+            },
+        },
+        handle));
+    return llfio::success();
+  }();
+
+  if (result.has_error()) {
+    return error_code(result.assume_error());
+  }
+
+  return 0;
+}
+
+extern "C" int blake3_hasher_update_mmap(blake3_hasher *self,
+                                         char const *path) noexcept {
+  bool use_tbb = false;
+  return blake3_hasher_update_mmap_base(self, path, use_tbb);
+}
+
+#if defined(BLAKE3_USE_TBB)
+extern "C" int blake3_hasher_update_mmap_tbb(blake3_hasher *self,
+                                             char const *path) noexcept {
+  bool use_tbb = true;
+  return blake3_hasher_update_mmap_base(self, path, use_tbb);
+}
+#endif // BLAKE3_USE_TBB
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -319,7 +319,7 @@ jobs:
 
   # CMake build test (Library only).
   cmake_current_build:
-    name: CMake ${{ matrix.os }} CC=${{ matrix.toolchain.cc }} CXX=${{ matrix.toolchain.cxx }} TBB=${{ matrix.use_tbb }}
+    name: CMake ${{ matrix.os }} CC=${{ matrix.toolchain.cc }} CXX=${{ matrix.toolchain.cxx }} LLFIO=${{ matrix.use_llfio }} TBB=${{ matrix.use_tbb }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -333,6 +333,7 @@ jobs:
           { cc: clang-cl, cxx: clang-cl },
           { cc: gcc, cxx: g++ },
         ]
+        use_llfio: [OFF, ON]
         use_tbb: [OFF, ON]
         exclude:
           - os: macOS-latest
@@ -346,6 +347,9 @@ jobs:
           - os: windows-latest
             toolchain: { cc: clang, cxx: clang++ }
             use_tbb: ON
+          - os: windows-latest
+            toolchain: { cc: gcc, cxx: g++ }
+            use_llfio: ON
           - os: windows-latest
             toolchain: { cc: gcc, cxx: g++ }
             use_tbb: ON
@@ -368,7 +372,7 @@ jobs:
       - name: CMake generation, build, install
         run: |
           ${{ matrix.os != 'windows-latest' || '& "C:/Program Files/Microsoft Visual Studio/2022/Enterprise/Common7/Tools/Launch-VsDevShell.ps1" -Arch amd64 -SkipAutomaticLocation' }}
-          cmake -S c -B c/build -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}/target -DCMAKE_C_COMPILER=${{ matrix.toolchain.cc }} -DCMAKE_CXX_COMPILER=${{ matrix.toolchain.cxx }} -DBLAKE3_USE_TBB=${{ matrix.use_tbb }} -DBLAKE3_FETCH_TBB=${{ matrix.os == 'windows-latest' && 'YES' || 'NO' }} -DBLAKE3_EXAMPLES=ON
+          cmake -S c -B c/build -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}/target -DCMAKE_C_COMPILER=${{ matrix.toolchain.cc }} -DCMAKE_CXX_COMPILER=${{ matrix.toolchain.cxx }} -DBLAKE3_USE_LLFIO=${{ matrix.use_llfio }} -DBLAKE3_FETCH_LLFIO=YES -DBLAKE3_USE_TBB=${{ matrix.use_tbb }} -DBLAKE3_FETCH_TBB=${{ matrix.os == 'windows-latest' && 'YES' || 'NO' }} -DBLAKE3_EXAMPLES=ON
           cmake --build c/build --target install
   cmake_3-9_build:
     name: CMake 3.9.6 ubuntu-latest
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -17,6 +17,12 @@ project(libblake3
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
+option(BLAKE3_USE_LLFIO "Enable llfio mmap support" OFF)
+set(BLAKE3_USE_LLFIO_DIR "" CACHE STRING "CMake subdirectory for existing llfio")
+set(BLAKE3_USE_LLFIO_TARGET "llfio_hl" CACHE STRING "Library target to use for llfio")
+option(BLAKE3_USE_LLFIO_EXPERIMENTAL_STATUS_CODE "Enable llfio experimental SG14 status code support" ON)
+option(BLAKE3_FETCH_LLFIO "Allow fetching llfio from GitHub if not found on system" OFF)
+
 option(BLAKE3_USE_TBB "Enable oneTBB parallelism" OFF)
 option(BLAKE3_FETCH_TBB "Allow fetching oneTBB from GitHub if not found on system" OFF)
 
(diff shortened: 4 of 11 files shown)


silvanshade:
There is no option to request reviewers and nobody is assigned so I'll just ask:

@BurningEnlightenment and @oconnor663 can you review this? Thanks.

c/blake3_llfio.cpp
```diff
@@ -0,0 +1,103 @@
...
+    return file;
+  } else {
+    auto &&map = llfio::mapped_file_handle(
+        std::move(file),                   // transfer ownership
+        0,                                 // reserve maximum extent
+        llfio::section_handle::flag::read, // map as read-only
+        0                                  // map from the start
+    );
+    return std::pair{std::move(map), file_size};
+  }
```
BurningEnlightenment, line 59: This is prone to TOCTOU, the file size might change between `file.maximum_extent()` and `llfio::mapped_file_handle()`.

c/blake3_llfio.cpp
```diff
@@ -0,0 +1,103 @@
...
+                         return llfio::success();
+                       },
+                   },
+                   mmap));
+    return llfio::success();
+  }();
+  if (result.has_error()) {
+    // Explicitly set errno on error since this doesn't always happen
+    // automatically on some platforms such as Windows.
+    errno = static_cast<int>(result.error().value().sc);
```
BurningEnlightenment, line 88: llfio leverages NT APIs on Windows, i.e. the status code might be a NTSTATUS and not a Win32 code. This is problematic due to the fact that NTSTATUS and Win32 codes have overlapping value ranges.
One solution would be to convert to HRESULT, but maybe there is a more elegant one. Needs additional thought.

Furthermore llfio might be built by the user with `std::error_code`, i.e. `.sc` will produce a compiler error.

And last but not least errno might be dirty, i.e. we should clear it on success or switch the return type from void to bool to indicate success.

c/blake3_llfio.cpp
```diff
@@ -0,0 +1,103 @@
...
+template <class... Ts> struct overloads : Ts... {
+  using Ts::operator()...;
+};
+// NOTE: deduction guide only needed for Apple Clang now.
+template <class... Ts> overloads(Ts...) -> overloads<Ts...>;
+
+INLINE auto copy_wide(blake3_hasher *self, llfio::path_view path,
+                      llfio::file_handle &file) noexcept
+    -> llfio::result<std::uint64_t> {
+  std::array<std::byte, 65535> buffer{};
```
BurningEnlightenment, line 19:
```suggestion
  std::array<std::byte, 65535> buffer;
```
I think this is a case which warrants uninitialized memory.

`#include <array>` is missing, alternatively switch to a plain old C-array.

c/blake3_llfio.cpp
```diff
@@ -0,0 +1,103 @@
...
+    // Mapping small files is not worth it, and some special files that can't be
+    // mapped report a size of zero.
+    return file;
+  } else {
+    auto &&map = llfio::mapped_file_handle(
+        std::move(file),                   // transfer ownership
+        0,                                 // reserve maximum extent
+        llfio::section_handle::flag::read, // map as read-only
+        0                                  // map from the start
+    );
```
BurningEnlightenment, line 57: Mapping the file to memory is not guaranteed to work. On 32bit platforms memory mapping is especially prone to failing due to address space exhaustion and/or fragmentation, i.e. this needs some proper error handling and a fallback to ordinary file reading.

c/blake3_llfio.cpp
```diff
@@ -0,0 +1,103 @@
...
+      return std::move(result).as_failure();
+    }
+    if (result.has_value()) {
+      auto bytes_read = result.value();
+      blake3_hasher_update(self, buffer.data(), bytes_read);
+      total += bytes_read;
+      if (bytes_read == 0) {
+        break;
+      }
+    }
```
BurningEnlightenment, line 34: Non-portable EOF handling see ned14/llfio#29

c/blake3_llfio.cpp
```diff
@@ -0,0 +1,103 @@
...
+  auto result = [=]() -> llfio::result<void> {
+    OUTCOME_TRY(auto file, llfio::file({}, path));
+    OUTCOME_TRY(auto mmap, maybe_mmap_file(std::move(file)));
+    OUTCOME_TRY(
+        std::visit(overloads{
+                       [=](std::pair<llfio::mapped_file_handle,
+                                     llfio::file_handle::extent_type> &pair)
+                           -> llfio::result<void> {
+                         blake3_hasher_update_base(self, pair.first.address(),
+                                                   pair.second, use_tbb);
```
BurningEnlightenment, line 74: This needs to be wrapped in a signal handler, use the [`signal_guard`] from quickcpplib (which is a llfio building block).

[`signal_guard`]: https://github.com/ned14/quickcpplib/blob/master/include/quickcpplib/signal_guard.hpp#L261

c/blake3_llfio.cpp
```diff
@@ -0,0 +1,103 @@
...
+  using Ts::operator()...;
+};
+// NOTE: deduction guide only needed for Apple Clang now.
+template <class... Ts> overloads(Ts...) -> overloads<Ts...>;
+
+INLINE auto copy_wide(blake3_hasher *self, llfio::path_view path,
+                      llfio::file_handle &file) noexcept
+    -> llfio::result<std::uint64_t> {
+  std::array<std::byte, 65535> buffer{};
+  std::uint64_t total = 0;
```
BurningEnlightenment, line 20: I don't think we use the value of `total` anywhere, right?

BurningEnlightenment (changes requested):
Thanks for submitting the PR!

I also noticed a few missing includes / reliance on transitive includes like `std::move` => `<utility>`, `OUTCOME_TRY` => `<outcome-basic.hpp>`, etc.

I haven't looked at the CMake files, yet.

I'll allocate some time next weekend to fix some of the issues.

c/blake3_llfio.cpp
silvanshade: Noted. Will fix this.

c/blake3_llfio.cpp
silvanshade: See here in the CMake configuration: https://github.com/silvanshade/BLAKE3/blob/fac71c9e6748e61e0c280e4e905f8f808b3c2ae8/c/dependencies/llfio/CMakeLists.txt#L27-L42

I'm specifically configuring the build with `LLFIO_EXPERIMENTAL_STATUS_CODE` in order to allow compiling llfio as a header-only library. The documentation states the following:

> LLFIO has your choice of header-only, static library, and shared library build modes. Note that on Microsoft Windows, the default header only configuration is unsafe to use outside of toy projects. You will get warnings of the form:
>
> [...]

> The cause is that <system_error> has a design flaw where custom error code categories are unsafe when shared libraries are in use.

> 
> You have **one of three choices** here: (i) Use [experimental SG14 status_code](https://wg21.link/P1028) which doesn't have this problem (define LLFIO_EXPERIMENTAL_STATUS_CODE=1) (ii) Use the NT kernel error category as a shared library ([see its documentation](https://github.com/ned14/ntkernel-error-category)) (iii) Don't use header only LLFIO on Windows (see below).

Of the three choices available, the header-only with `LLFIO_EXPERIMENTAL_STATUS_CODE` seems the cleanest and most portable.

I couldn't find any vendor-provided packages available for llfio for any OS, so relying on it being available pre-compiled doesn't seem like a viable option.

I'd also prefer to avoid compiling it in-tree as a shared lib on Windows if we don't need that.

> Furthermore llfio might be built by the user with std::error_code, i.e. .sc will produce a compiler error.

Since llfio is configured as a private dependency and header-only library, I don't see how this can be a problem. Our use of llfio does not interfere with other already existing uses of it elsewhere in the user's code base.

> And last but not least errno might be dirty, i.e. we should clear it on success or switch the return type from void to bool to indicate success.

Good point. I would prefer to return a bool rather than set-on-success since it is cleaner and more informative. I will change that.

c/blake3_llfio.cpp
silvanshade: Noted. I will change this.

c/blake3_llfio.cpp
silvanshade: I think I did it this way because I was under the impression that if the memory mapping failed, the error would propagate through when you tried to do a read or some other operation.

But the point about catching it on low memory devices and responding earlier with a different strategy is a good idea.

I'll refactor this part to handle it differently.

c/blake3_llfio.cpp
silvanshade: I did see that issue but didn't read the whole thing in detail beyond the first couple comments. I see there's more subtlety involved there so I'll refactor this.

c/blake3_llfio.cpp
silvanshade: We use it here: https://github.com/BLAKE3-team/BLAKE3/blob/fac71c9e6748e61e0c280e4e905f8f808b3c2ae8/c/blake3_llfio.cpp#L30

and here:
https://github.com/BLAKE3-team/BLAKE3/blob/fac71c9e6748e61e0c280e4e905f8f808b3c2ae8/c/blake3_llfio.cpp#L36

BurningEnlightenment: yes, I should've elaborated more. We don't do anything with the return value of `copy_wide` at the call-sites and therefore we do nothing with the `total` value AFAICT

c/blake3_llfio.cpp
BurningEnlightenment: > Since llfio is configured as a private dependency and header-only library, I don't see how this can be a problem. As far as I understand, our use of llfio should not interfere with other already existing uses of it elsewhere in the user's code base.

Sadly, this is not the case. Formally speaking: Using llfio with different configurations in the same project results in One Definition Rule (ODR) violations for basically every entity. ODR violations make the whole program "ill-formed no diagnostic required" (IFNDR), it basically stops being a C++ program from the C++ standard POV.

The practical consequences can be illustrated with the following example: We build blake3 as a static library and llfio in its header variant is used in two configurations like `std::error_code` and `status_code` (progam's vs blake3's). All `inline` functions will be emitted by the compiler into the object file and the linker is allowed to deduplicate them by picking any of them. This means that either our program or the blake3 library will not get ABI incompatible return values.

> I couldn't find any vendor-provided packages available for llfio for any OS, so relying on it being available pre-compiled doesn't seem like a viable option.

Indeed, OS packagers are a bad source for C++ libraries (though llfio has been added to the FreeBSD ports tree IIRC). Typically you would use a C++ package manager like `vcpkg` or `conan`.

btw. I also maintain llfio and its buildsystem and I'm therefore vaguely familiar with it :wink:

silvanshade: > The practical consequences can be illustrated with the following example: We build blake3 as a static library and llfio in its header variant is used in two configurations like `std::error_code` and `status_code` (progam's vs blake3's). All `inline` functions will be emitted by the compiler into the object file and the linker is allowed to deduplicate them by picking any of them. This means that either our program or the blake3 library will not get ABI incompatible return values.

Okay, I see how this can be a problem.

> Indeed, OS packagers are a bad source for C++ libraries (though llfio has been added to the FreeBSD ports tree IIRC). Typically you would use a C++ package manager like `vcpkg` or `conan`.
>
> btw. I also maintain llfio and its buildsystem and I'm therefore vaguely familiar with it 😉

I have some familiarity with `vcpkg` (not `conan` though), but to be more specific, I wasn't sure how we should detect local availability of `llfio` regardless of the source.

If I remember correctly, `llfio` didn't seem to be set up with a CMake find module definition so we could just use `find_package`. This is why I checked for the `llfio::hl` target specifically (after loading from `FetchContent`).

How would you propose we attempt to detect and use a locally available `llfio`?

One strategy I am imagining is that if `BLAKE3_USE_LLFIO` is set, we check for a (user-defined) value for `LLFIO_LIBRARY` and if defined we attempt to load the value as subdirectory with CMake (which we expect to define the `llfio::*` targets), otherwise if `BLAKE3_FETCH_LLFIO` is set, use `FetchContent` and configure like we have it now, otherwise print an error message about missing `llfio`.

c/blake3_llfio.cpp
silvanshade: I wrote it that way because I was trying to follow the structure of the definitions in [`io.rs`](https://github.com/BLAKE3-team/BLAKE3/blob/master/src/io.rs) and there `copy_wide` is written to return `total`.

But you're right it isn't actually used for anything here nor is it actually used in equivalent the Rust code.

I could remove it.

c/blake3_llfio.cpp
oconnor663, line 19: It's actually important for performance that the buffer size is a power of 2. Can we use 65536 for the size here, as in the Rust code?

c/blake3_llfio.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+namespace llfio = LLFIO_V2_NAMESPACE;
+
+template <class... Ts> struct overloads : Ts... {
+  using Ts::operator()...;
+};
+// NOTE: deduction guide only needed for Apple Clang now.
+template <class... Ts> overloads(Ts...) -> overloads<Ts...>;
+
+INLINE auto copy_wide(blake3_hasher *self, llfio::path_view path,
```
oconnor663, line 16: Is this `path` parameter unused?

silvanshade: Yes, it's unused. I think that must have been from a refactoring and forgot to remove it.

c/blake3.h
```diff
@@ -69,6 +75,18 @@ BLAKE3_API void blake3_hasher_init_derive_key_raw(blake3_hasher *self, const voi
...
+#if defined(BLAKE3_USE_LLFIO)
+BLAKE3_API void blake3_hasher_update_mmap(blake3_hasher *self,
+                                          const char *path) BLAKE3_NOEXCEPT;
+#endif // BLAKE3_USE_LLFIO
+#if defined(BLAKE3_USE_TBB)
+BLAKE3_API void blake3_hasher_update_tbb(blake3_hasher *self, const void *input,
+                                         size_t input_len);
+#if defined(BLAKE3_USE_LLFIO)
+BLAKE3_API void blake3_hasher_update_mmap_tbb(blake3_hasher *self,
+                                              const char *path) BLAKE3_NOEXCEPT;
```
oconnor663, line 87: `blake3_hasher_update_mmap` and `blake3_hasher_update_mmap_tbb` should probably be able to report IO errors. Is it possible to get an `errno`-compatible `int` here?

oconnor663: Also what's the right way to express a path in a C API that allows for UTF-16 paths on Windows?

silvanshade: We could do that.

The way it's handled right now is that `blake3_hasher_update_mmap_base` will set `errno` if an error occurs: https://github.com/BLAKE3-team/BLAKE3/blob/fac71c9e6748e61e0c280e4e905f8f808b3c2ae8/c/blake3_llfio.cpp#L88

In the example I check for that by clearing `errno` before the call and checking it afterwards: https://github.com/BLAKE3-team/BLAKE3/blob/fac71c9e6748e61e0c280e4e905f8f808b3c2ae8/c/example-mmap.c#L25-L27

BurningEnlightenment commented earlier that we could return a bool to be more informative regarding whether an error occurred (so it's not necessary to do the clear manually first).

But an alternative would be to just return the code as an `int`. Is that what you would prefer?

silvanshade: > Also what's the right way to express a path in a C API that allows for UTF-16 paths on Windows? Would we have to take some monstrosity like this?
> 
> ```
> typedef struct {
>     uint16_t* utf16_path;  // Windows
>     char* utf8_path;       // POSIX
> } FilePath;
> ```

Something like that I think, but if we want to be rigorous, we'd probably need to use `wchar_t` for Windows and `char8_t` for POSIX. But `char8_t` requires C23.

c/blake3_llfio.cpp
```diff
@@ -0,0 +1,103 @@
...
+void blake3_hasher_update_mmap(blake3_hasher *self, char const *path) noexcept {
+  bool use_tbb = false;
+  blake3_hasher_update_mmap_base(self, path, use_tbb);
+}
+
+#if defined(BLAKE3_USE_TBB)
+void blake3_hasher_update_mmap_tbb(blake3_hasher *self,
+                                   char const *path) noexcept {
+  bool use_tbb = true;
+  blake3_hasher_update_mmap_base(self, path, use_tbb);
```
oconnor663, line 101: I think both `blake3_hasher_update_mmap` and `blake3_hasher_update_mmap` need to be `extern "C"`.

c/blake3.h
oconnor663: > But an alternative would be to just return the code as an int. Is that what you would prefer?

Oh nice, yeah that sounds great.

oconnor663:
Could you give me the high level tradeoffs for LLFIO vs [`mio`](https://github.com/vimpunk/mio) vs Boost [`mapped_file`](https://live.boost.org/doc/libs/1_87_0/libs/iostreams/doc/classes/mapped_file.html)? I guess another option we could consider is providing only the TBB side of the equation (since that necessarily touches BLAKE3 internaly) but leaving mmap'ing to the caller?

I was playing around with adding support for this feature to the `build.rs` file in the `blake3_c_rust_bindings` test harness, and looks like pulling in these headers means cloning three different git repos plus their submodules. I'm willing to do that (our Rust code has a gajillion transitive dependencies if you count dev/test deps), but if one of the alternatives is lighter weight that might be nice? I acknowledge that using a Rust test harness to run C/C++ code is a wacky choice :-D

silvanshade:
> Could you give me the high level tradeoffs for LLFIO vs [`mio`](https://github.com/vimpunk/mio) vs Boost [`mapped_file`](https://live.boost.org/doc/libs/1_87_0/libs/iostreams/doc/classes/mapped_file.html)? I guess another option we could consider is providing only the TBB side of the equation (since that necessarily touches BLAKE3 internaly) but leaving mmap'ing to the caller?

@oconnor663 Sorry I missed this earlier but I see you merged the TBB PR. I think that's fine; it's still an improvement to have to multi-threading for C even if the caller has to figure out the memory-mapping for now.

I don't have much familiarity with `mio` but there was a reddit thread about it [here](https://www.reddit.com/r/cpp/comments/9p4ixb/mio_crossplatform_headeronly_c11_library_for/).

The llfio author (ned14) [commented](https://www.reddit.com/r/cpp/comments/9p4ixb/comment/e7zn6np/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button):

>From a quick scan through his sources:
>
>LLFIO is far more future standards focused, which makes sense as the reference implementation for http://wg21.link/P1031 Low level file i/o. So specifically, he's got lots of useful iterator support and so on. LLFIO goes no further than giving you a span<T>, it assumes you'll be using Ranges on those. LLFIO also uses Outcome as a standin for deterministic exceptions with the proposed std::error. He's using C++ 11 error_code, which is reasonable.
>
>LLFIO does a ton load more stuff, and is correspondingly more complex. You get non-mmapped file i/o, async i/o, ability to polymorphically interchange i/o backends, various generic algorithms for excluding other users from parts of files, or many files, and so on.
>
>LLFIO implements a full fat virtual memory control layer, so you can kick in or out individual pages, use large pages, remap pages, and so on.
>
>LLFIO implements a full filesystem layer, so you get path views (proposed for standardisation), ability to enumerate directories correctly rather than the abomination which is directory_iterator, symlinks and so on.
>
>Finally LLFIO handles are stackable, not that that is in the current reference library. I'm currently working on the first stackable handle type, it implements crypto random XOR.
>
>His library looks well implemented, though he could do with more comprehensive unit testing. I wouldn't have any qualms in recommending its use.

The `mio` library seems to be less actively developed now.

Regarding LLFIO vs Boost (`mapped_file` and other features), it's my impression that LLFIO was developed partly in response to needing more complete or more advanced or more performant functionality than what was available in Boost.

I believe the intention was also that parts of LLFIO were to be proposed for addition to the C++ standard (more on this later) whereas Boost has often kind of evolved its own way of doing this for various reasons. I would expect LLFIO to be more forward looking in its design.

The main reason I chose LLFIO is because it was actively developed and it looked like parts of it were likely to be added to the standard. But I recently saw this [comment](https://github.com/ned14/llfio/issues/127#issuecomment-2186241660) and it sounds like the author is stepping away from the C++ WG stuff and deprioritizing C++ projects.

I'm less sure about whether it's worth the effort to get LLFIO-based memory-mapping to work for the C lib now. 

Maybe @BurningEnlightenment has some thoughts about all of this.

> I was playing around with adding support for this feature to the `build.rs` file in the `blake3_c_rust_bindings` test harness, and looks like pulling in these headers means cloning three different git repos plus their submodules. I'm willing to do that (our Rust code has a gajillion transitive dependencies if you count dev/test deps), but if one of the alternatives is lighter weight that might be nice? I acknowledge that using a Rust test harness to run C/C++ code is a wacky choice :-D

Yeah, I don't know really. This sounds like maybe more trouble than its worth.

If you do everything through CMake and use `FetchContent` and build LLFIO as header-only, it's pretty smooth. But I'm finding that almost any other configuration (especially on Windows) is a lot more complex than I anticipated, even for the C++ side.

I have resolved most of the feedback and plan to update the PR still but maybe there's a better option or maybe it's fine to just not have memory-mapping.

Perhaps a much simpler approach would be to split the memory-mapping code from Rust into a separate crate which could be statically linked to the C library.

c/blake3_llfio.cpp
silvanshade: > > Furthermore llfio might be built by the user with std::error_code, i.e. .sc will produce a compiler error.
> 
> Given the above, how can we detect which configuration is in use in order to handle it appropriately in the different cases? Can you give an example?

I tried to generalize the implementation to work with either `std::error_code` or `status_code` and also to allow specifying whether to use an existing `llfio` or to fetch and whether to build as header-only (the default) or static or shared.

Unfortunately, I can't seem to get most of the non-header-only configurations to even compile on Windows. I haven't tested that on other platforms yet.

You can try yourself by specifying, for example, `-DBLAKE3_USE_LLFIO=1 -DBLAKE3_USE_LLFIO_TARGET=llfio_dl -DBLAKE3_USE_LLFIO_EXPERIMENTAL_STATUS_CODE=OFF -DBLAKE3_FETCH_LLFIO=1`.

For non-header-only builds on Windows, the compile fails with several unresolved symbols unless exceptions are enabled, but as far as I understand this should be supported according to https://github.com/ned14/llfio/issues/127.

Another problem is that if `EXCLUDE_FROM_ALL` is specified in:

```cmake
FetchContent_Declare(
  llfio
  GIT_REPOSITORY https://github.com/ned14/llfio
  GIT_TAG 20250206
  GIT_SHALLOW TRUE
  EXCLUDE_FROM_ALL
)
```

or:

```cmake
add_subdirectory(${BLAKE3_USE_LLFIO_DIR} EXCLUDE_FROM_ALL)
```

then the `llfio` libs won't be compiled so the `example-mmap` will just fail when run because the libraries are missing. I would have thought that adding the `llfio_dl` to the target would force these to compile.

oconnor663:
I've added `example_tbb.c` (https://github.com/BLAKE3-team/BLAKE3/commit/2374b56df6be47d2865e512dc80c1f4f4c0b89d3) as an example of using `mmap` directly in C (so not supporting Windows). Maybe that's sufficient? Let me know what you think, if you get a chance to look at it.

sneves:
The handles created by `open` are never `close`d in the example.

silvanshade:
> I've added `example_tbb.c` ([2374b56](https://github.com/BLAKE3-team/BLAKE3/commit/2374b56df6be47d2865e512dc80c1f4f4c0b89d3)) as an example of using `mmap` directly in C (so not supporting Windows). Maybe that's sufficient? Let me know what you think, if you get a chance to look at it.

Cool, that seems sufficient I think.

The motivating use-case for this PR was to enable memory-mapped IO for the BLAKE3 hashing in nix. But nix is already depending on Boost locally so it's probably simpler to just use `Boost::mapped_file` there.

I'll go ahead and close this then. Thanks for the feedback.

oconnor663:
Cool, thanks for all the work you've done to get this feature in.

> The handles created by open are never closed in the example.

Woops, fixed.

PR #447, opened 2025-02-12
fix build.rs errors on win7 targets

OmerCheri:
When trying to compile on the x86_64-win7-windows-msvc target, we're getting the following error:
It seems that the missing string "pc" is causing the build script to think we're building for Unix.

warning: blake3@1.4.1: The C compiler "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Tools\\MSVC\\14.29.30133\\bin\\HostX64\\x64\\cl.exe" does not support -mavx512f and -mavx512vl.
warning: blake3@1.4.1: cl : Command line warning D9002 : ignoring unknown option '-std=c11'
warning: blake3@1.4.1: cl : Command line warning D9024 : unrecognized source file type 'c/blake3_sse2_x86-64_unix.S', object file assumed
warning: blake3@1.4.1: cl : Command line warning D9027 : source file 'c/blake3_sse2_x86-64_unix.S' ignored
  CFLAGS = None
  running: "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Tools\\MSVC\\14.29.30133\\bin\\HostX64\\x64\\cl.exe" "-nologo" "-MT" "-O1" "-Z7" "-Brepro" "-W4" "-std=c11" "-FoC:\\Users\\ContainerAdministrator\\AppData\\Local\\Temp\\66ea281f-7371-429e-8787-507757fb1b83\\x86_64-win7-windows-msvc\\release\\build\\blake3-15c154c91f03ef71\\out\\c/blake3_sse2_x86-64_unix.o" "-c" "c/blake3_sse2_x86-64_unix.S"
  cargo:warning=cl : Command line warning D9002 : ignoring unknown option '-std=c11'

  cargo:warning=cl : Command line warning D9024 : unrecognized source file type 'c/blake3_sse2_x86-64_unix.S', object file assumed

  cargo:warning=cl : Command line warning D9027 : source file 'c/blake3_sse2_x86-64_unix.S' ignored

  cl : Command line warning D9021 : no action performed
  exit code: 0
[...]

PR #448, opened 2025-02-16
Add `Hash::from_slice` to handle conversion from `&[u8]` bytes

joshtriplett:
This is a convenience method, to avoid having to first call
`<[u8; 32]>::try_from(slice)?`.

I've run into the need for this many times, typically when deserializing, such
as from a database.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -236,6 +236,13 @@ impl Hash {
         Self(bytes)
     }
 
+    /// Create a `Hash` from its raw bytes representation as a slice.
+    ///
+    /// Returns an error if the slice is not exactly 32 bytes long.
+    pub fn from_slice(bytes: &[u8]) -> Result<Self, core::array::TryFromSliceError> {
+        Ok(Self::from_bytes(bytes.try_into()?))
+    }
+
     /// Encode a `Hash` in lowercase hexadecimal.
     ///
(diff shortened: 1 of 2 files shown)


oconnor663:
Seems reasonable. Any reason not to add a `TryFrom` impl at the same time?

oconnor663:
(For example witih this commit on top of yours: https://github.com/BLAKE3-team/BLAKE3/tree/from-slice-with-tryfrom)

joshtriplett:
> Seems reasonable. Any reason not to add a `TryFrom` impl at the same time?

I think there is a reason to at least consider whether or not to add one. `TryFrom` impls are often used more implicitly and with less checking of documentation, and I'd be mildly concerned about people mixing up hex representation vs binary. That's not a danger with the existing `From` because it comes from `[u8; 32]`, whose length means it must be binary. But with a `TryFrom` impl for `&[u8]`, it would be easier to accidentally call it on the bytes of a hex string you just read in.

That said, happy to add it if you think that's worth the tradeoff.

oconnor663:
Good point, I forgot that `from_hex` actually takes `impl AsRef<[u8]>`. Ok, let's stick with this for now. I need to ship https://github.com/BLAKE3-team/BLAKE3/pull/447, so I'll go ahead and push out a release right now :)

PR #450, opened 2025-02-18
build(cmake): improve CMake backwards compatibility 

BurningEnlightenment:
There are no default paths for target installs on CMake 3.9.

Test the minimum required CMake version on CI. We only test on Ubuntu due to the fact that CMake 3.9.6 does not
recognize modern MSVC versions and the main use case for these ancient
CMake versions are extremely conservative Linux distributions.


BurningEnlightenment:
@oconnor663 if nothing comes up, I'll go ahead and merge this on sunday.

oconnor663:
No objections :)

PR #454, opened 2025-03-14
Include nix under Adoption & deployment

silvanshade:


oconnor663:
Thank you for doing the upstream integration :)

(Looks like one of our dependencies did an MSRV bump that's failing CI. I'll deal with that separately. Might be worth refactoring things slightly to only check the MSRV against `cargo build`.)

PR #455, opened 2025-03-15
[C library] use Windows asm for cygwin

ognevny:
in cygwin falling to Unix asm make build fails with it

cc https://github.com/msys2/MSYS2-packages/pull/5260


BurningEnlightenment (approved):
*sigh* LGTM

Issue #229, opened 2022-03-04
C and multithreading


odiferousmint:
Hello. I just read `C/README.md`.

Why not add optional support for multithreading? Perhaps it can be a configuration flag, say, `--enable-multithreading=openmp` (default being `openmp`, which may be the only implementation for a while, but would allow flexibility in case someone implements it through different ways).

This was people who want to implement it themselves can just compile it without multithreading.

Alternatively, we may provide functions that are multithreaded and ones that are not.

oconnor663:
The compile-time and runtime dependencies of OpenMP are my biggest concern. I'm sure everything could be made to work on common platforms, but I worry that the long tail of Windows support and Debian 9 and weird old compilers and and who-knows-what else will be a burden, and that an OpenMP implementation will never match the portability of our existing code. We already do some nontrivial things with CPU feature detection, for example, but we're able to encapsulate that well enough that callers don't need to know about it. If there was way to accomplish something similar for threading, e.g. by vendoring and statically linking some open source library, I would be very interested in looking into it.

My second biggest is that I don't have any experience with OpenMP, coupled with how multithreading in C is inherently unsafe. The API requirements for multithreading BLAKE3 are quite modest, just some flavor of `fork` and `join`. But there's some risk that I'll put a memory barrier in the wrong place and cause some subtle race condition that only manifests on e.g. ARMv7. I'd feel better about this if someone who has non-zero experience with OpenMP wrote the code.

These are all surmountable problems, and having the working Rayon-based Rust implementation as a point of comparison makes things easier. (Though it also raises the possibility of just calling the Rust code from C, if you're willing to tolerate beefy dependencies.) But to me these problems don't...spark joy.

oconnor663:
See also https://github.com/BLAKE3-team/BLAKE3/issues/184.

odiferousmint:
> OpenMP implementation will never match the portability of our existing code

That may be true, although I suggest you take a look at https://www.openmp.org/resources/openmp-compilers-tools/, for one!

> My current priority (apart from teaching at NYU) is on fixing the poor performance of b3sum on spinning disks, by implementing a different multithreading strategy.

Are you still working on it? I am using a HDD. I can attempt at making it multithreaded (with OpenMP or whatever you'd rather prefer), but cannot test on ARMv7, nor do I know much about it, unfortunately. I do know some, as I have done some serious file locking stuff before, and there were complications at all levels, be it hardware, the filesystem, the kernel, etc.

[...]

Issue #184, opened 2021-07-26
Use case for multithreaded hashing in C


fcorbelli:
From a quick C/ (++) implementation of BLAKE3, in my ZPAQ's fork archiver, I found modest performance (file hashing), similar to ~ SHA256 (~300MB/s) on single thread (just C code, no ASM).

Indeed good for a cryptographic hash, but the inability to use multithreading (which in my opinion is the real advantage of BLAKE3) leaves me unsatisfied.

The use is twofold: integrity check of archived files and deduplication of documents from filesystems

Are there any plans for this (multithreaded in C/C ++) via pthread or whatever?

Thank you for all reply

oconnor663:
May I recommend benchmarking BLAKE3 with the assembly implementations? They are several times faster than the portable code on most x86 machines :)

It's unlikely we'll implement our own work-stealing multithreading strategy on top of pthreads. If we were going to add this into the C implementation, we'd probably use OpenMP. I don't have much experience with OpenMP myself, but my understanding is that it's a nontrivial dependency that not everyone would be happy to have in their build. (For example, it seems like the MSVC implementation of OpenMP targets a version of the standard that's almost 20 years old?)

It might make sense to ask, would it be cleaner to just expose C bindings for the Rust crate and link against that? I don't know of anyone doing this in production, but I was able to whip up an example just now: https://github.com/oconnor663/rust_blake3_c_bindings

fcorbelli:
Thanks for the reply.   

@assembly. In fact no, I want/need to compile on different platforms, sometimes even Annapurna (QNAP NAS). I am making my naive benchmarker (for my use case, of course) to compare against others crypto and non cryptographically strong hasesh more easily. 

@multuthread: I do not like OpenMP too (neither pthread), but pthread is just about diffuse to work on Windows/Linux/FreeBSD without much hassle. Not perfect, but... it works. In fact does NOT work for C++, but BLAKE3 is C. No ifdef, no compiler-time-checks etc  OpenMP is just like a kudzu.  
@Rust. No for the very same reason:  want/need a single .cpp file that will compile without libraries, linking, strange assembly, make files, any kind of #define or whatever "tricks".

oconnor663:
[...]

PR #457, opened 2025-03-19
Implement OpenMP-based parallelism for C lib

ibmibmibm:
Mimics all TBB changes (#445).

Related issues: #229 #184

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -459,3 +459,23 @@ jobs:
     - name: cargo test C bindings with TBB
       run: cargo test --features=tbb
       working-directory: ./c/blake3_c_rust_bindings
+
+  openmp_rust_bindings_test:
+    name: OpenMP test bindings ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest", "macOS-latest"]
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@stable
+    - name: install libomp
+      if: matrix.os == 'macOS-latest'
+      run: |
+        brew install libomp
+        echo "CFLAGS=-I$(brew --prefix libomp)/include $CFLAGS" >> $GITHUB_ENV
+        echo "RUSTFLAGS=-L$(brew --prefix libomp)/lib $RUSTFLAGS" >> $GITHUB_ENV
+    - name: cargo test C bindings with OpenMP
+      run: cargo test --features=openmp
+      working-directory: ./c/blake3_c_rust_bindings
--- a/c/.gitignore
+++ b/c/.gitignore
@@ -1,6 +1,7 @@
 blake3
 example
 example_tbb
+example_openmp
 build/
 *.o
 
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -19,6 +19,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 option(BLAKE3_USE_TBB "Enable oneTBB parallelism" OFF)
 option(BLAKE3_FETCH_TBB "Allow fetching oneTBB from GitHub if not found on system" OFF)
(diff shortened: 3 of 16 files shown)


oconnor663:
I don't _love_ the prospect of maintaining two separate implementations of this, but I'm curious to hear your thoughts about how many callers might prefer/demand to use OpenMP instead of TBB.

Separately, there was some discussion in the TBB PR about how OpenMP might not perform well with the nested/dynamic task spawning strategy that we're using here. @sylvanshade [commented](https://github.com/BLAKE3-team/BLAKE3/pull/445#issuecomment-2649486407) that:

> ...OpenMP is generally not suitable for nested parallelism, which is exactly how the problem is being solved here with Rayon and the oneTBB implementation. Most of the performance recommendations I see regarding OpenMP suggest to avoid nested parallelism at all costs because it's known to be inefficient.

@sneves made some [performance measurements](https://github.com/BLAKE3-team/BLAKE3/pull/445#issuecomment-2649318813) that seemed to corroborate @sylvanshade's comment. And when I run the benchmarks you added to `blake3_c_rust_bindings` (thank you for being so thorough!), it looks like OpenMP is having some trouble:

```
$ cd c/blake3_c_rust_bindings && cargo +nightly bench --features=tbb,openmp 1024_kib
...
test bench_incremental_1024_kib        ... bench:     117,774.41 ns/iter (+/- 1,743.86) = 8903 MB/s
test bench_openmp_1024_kib             ... bench:     946,225.39 ns/iter (+/- 507,769.54) = 1108 MB/s
test bench_tbb_1024_kib                ... bench:      47,784.35 ns/iter (+/- 6,406.28) = 21944 MB/s
```

Have you seen anything different in your measurements? Are there other cases we should be testing?

cosmicexplorer:
See discussion at https://github.com/uxlfoundation/oneTBB/discussions/1204#discussioncomment-16527225. For the spack package manager, I would absolutely prefer an OpenMP fallback if at all possible. The comparison being made here is extremely unserious--oneTBB refuses to clarify its support for non-Intel processors, while OpenMP has been standardized for decades and is taught to undergraduates.

# on "maintenance"

I'm further confused by this comment @oconnor663:
> I don't love the prospect of maintaining two separate implementations of this

It's not clear that you're "maintaining" a single implementation of this, given that `c/README.md` describes a gcc compile command and there's absolutely no discussion of the hardware compatibility. I can't even tell what processor your CI is running on. There's a load of cross-compile tests but I can't see what actual processor is being used for each worker, and that's a pretty significant distinction if you want to use words like "maintaining" and "support".

## what "maintenance" would entail

It's quite possible that:
- oneTBB is well-known to work with non-Intel `x86{,_64}` processors,
- oneTBB is easy to build reliably in arbitrary environments.

There is no indication of this in `c/README.md`, and furthermore no real indication of specific hardware support that is being "maintained". I cannot find a discussion of hardware support for x86 CPUs in the PR history. Perhaps I've missed a discussion.

# why precise guarantees of hardware support are significant [bootstrapping]

I am very interested in using BLAKE3 as part of a bootstrapping package manager resistant to length extension of `.tar.zst` archives. I am also interested in evaluating how tree hashing may correspond to an external archive format somewhat like zip files, and more generally the application of tree hashing in a filesystem layer, since no filesystem makes use of the tree hashing features of BLAKE2 or BLAKE3 for some reason.

For these purposes, I need to be able to understand what platforms BLAKE3 is known to work on, and in particular which platforms its maintainers are interested in supporting. OpenMP makes these clear. oneTBB does not.

## special note on `rustc` and `cargo`
Rust is especially problematic because `rustc` requires depending upon an opaque binary downloaded from microsoft servers, and `cargo` is especially difficult to integrate into other build systems (see my rejected NGI Zero app attempting to fix this: https://circumstances.run/@hipsterelectron/114610077000401178). So I would really like to at least have the C impl to compare against.

# future work
At this point, it seems likely that my needs will not be served by the reference implementation, which is kind of fun because it gives me an excuse to implement this and learn more about the work (I deeply admire the BLAKE3 paper and consider it one of the most exciting cryptographic developments of the past decade).

Hopefully the above helps to describe what use cases would motivate using something like OpenMP and more generally the utility of portability to the goals of this codebase.

PR #458, opened 2025-03-27
add the `hazmat` module

oconnor663:
cc @rklaehn @qti3e

The main goal of this new API is to make it so that projects like Bao don't depend on [undocumented APIs](https://github.com/BLAKE3-team/BLAKE3/blob/83a60d9da6a44b719e258926e568c203d3b4dd36/src/guts.rs), and projects like Iroh don't need to maintain [their own BLAKE3 forks](https://github.com/n0-computer/iroh-blake3).

Some questions in my mind:
- Do we like the `Mode` enum that the merge subtrees functions take? Should it hold keys by value instead of by reference?
- The `max_subtree_len` and `left_subtree_len` functions aren't really necessary (most callers don't need the former, and the latter is easily replicated), but I included them because their doc comments seemed like the best place to talk about the rules for the tree structure. Is that valuable? Are there other things we should document?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -871,12 +871,12 @@ fn compress_subtree_to_parent_node<J: join::Join>(
 
 // Hash a complete input all at once. Unlike compress_subtree_wide() and
 // compress_subtree_to_parent_node(), this function handles the 1 chunk case.
-fn hash_all_at_once<J: join::Join>(input: &[u8], key: &CVWords, flags: u8) -> Output {
+fn hash_all_at_once<J: join::Join>(input: &[u8], key: &CVWords, flags: u8, counter: u64) -> Output {
     let platform = Platform::detect();
 
     // If the whole subtree is one chunk, hash it directly with a ChunkState.
     if input.len() <= CHUNK_LEN {
-        return ChunkState::new(key, 0, flags, platform)
+        return ChunkState::new(key, counter, flags, platform)
             .update(input)
             .output();
     }
@@ -887,7 +887,7 @@ fn hash_all_at_once<J: join::Join>(input: &[u8], key: &CVWords, flags: u8) -> Ou
         input_chaining_value: *key,
         block: compress_subtree_to_parent_node::<J>(input, key, 0, flags, platform),
         block_len: BLOCK_LEN as u8,
-        counter: 0,
+        counter,
         flags: flags | PARENT,
         platform,
     }
@@ -913,7 +913,7 @@ fn hash_all_at_once<J: join::Join>(input: &[u8], key: &CVWords, flags: u8) -> Ou
 /// This function is always single-threaded. For multithreading support, see
 /// [`Hasher::update_rayon`](struct.Hasher.html#method.update_rayon).
 pub fn hash(input: &[u8]) -> Hash {
-    hash_all_at_once::<join::SerialJoin>(input, IV, 0).root_hash()
+    hash_all_at_once::<join::SerialJoin>(input, IV, 0, 0).root_hash()
 }
 
 /// The keyed hash function.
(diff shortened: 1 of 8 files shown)


devinrsmith:
I'm interested in giving this a pass, as I currently use guts and implement my own left_subtree_len for various traversal logic.

rklaehn:
I think @redsolver might also be interested in this.

qti3e:
This is a nice API, thank you. I think it covers enough for me to not need a fork anymore. 

Maybe something worth documenting somewhere is how to find the index of a chunk in the post-order traversal of the tree. Which I usually find useful every time I have to deal with caching the internal leafs in their stack order.

```
   6
 2   5
0 1 3 4
-------
0 1 2 3

tree_index(0) -> 0
tree_index(1) -> 1
tree_index(2) -> 3
tree_index(3) -> 4
```

```rs
/// Compute the index of the n-th leaf in the array representation of the tree.
/// see: <https://oeis.org/A005187>
#[inline(always)]
pub const fn tree_index(block_counter: usize) -> usize {
    2 * block_counter - block_counter.count_ones() as usize
}


/// Inverse of [`tree_index`].
pub const fn block_counter_from_tree_index(index: usize) -> Option<usize> {
    let mut result = 0;
    let mut x = index;
    // tree_index(7)=11 ; 7 = 4+2+1
    // 11 = 4 + (2*4-1)
    // 4  = 1 + (2*2-1)
    // 1  = 0 + (2*1-1)
    //
    // f(11) = f(4) + 4
    // f(4) = f(1) + 2
    // f(1) = f(0) + 1
    // f(0) = 0
    //
    // f(n) = f(r) + 2^e(n) such that: n = r + (2*2^e(n) - 1) and 0<=r<=2^e(n)
    // solving for e(n) we have e(n)=floor(log2((n + 1) / 2))=floor(log2(n + 1)) - 1
    while x != 0 {
        // compute the number of elements in the largest left subtree that it is
        // itself a full power of 2 elements.
        // =floor(log2(x + 1))
        let exp = usize::BITS - (x + 1).leading_zeros() - 1;
        let m = 1usize << (exp - 1); // 2^(exp - 1)
                                     // find our position in the right subtree and perform
                                     // the next step recursivly.
        x -= 2 * m - 1;
        // our index is at least as largest as all of the items in our left
        // subtree. All of the `m`s we are adding are unique power of twos
        // so we can use the `|=` instead which looks cooler.
        result |= m;
    }
    if tree_index(result) == index {
        Some(result)
    } else {
        None
    }
}
```

rklaehn:
I have confirmed in https://github.com/n0-computer/bao-tree that this API works for us (iroh).

Regarding the Mode enum. I like it. I also don't mind the lifetime, although I doubt it matters very much since this is going to be inlined and optimized away anyway.

Regarding helpful bit twiddling functions like the above. I think there is a value in having them somewhere, but I am not sure if the hazmat API is the right place. The two fns left_subtree_len and max_subtree_len are an exception since they are helpful for ensuring correct usage of the guts API.

rklaehn:
I can't comment that much on the _xof functions, since we don't use them in bao-tree. It seems fine to me, but I am not a cryptographer.

Maybe @dignifiedquire or @matheus23 have feedback?

Otherwise I would say :shipit: ...

src/lib.rs
```diff
@@ -885,7 +884,7 @@ fn hash_all_at_once<J: join::Join>(input: &[u8], key: &CVWords, flags: u8) -> Ou
     // compress_subtree_to_parent_node().
     Output {
         input_chaining_value: *key,
-        block: compress_subtree_to_parent_node::<J>(input, key, 0, flags, platform),
+        block: compress_subtree_to_parent_node::<J>(input, key, counter, flags, platform),
         block_len: BLOCK_LEN as u8,
         counter: 0,
```
matheus23, line 889: Shouldn't this be `counter` instead of `0`?

matheus23: It's weird that this function is non-public and as far as I can tell, never called with `counter` being non-zero.

Perhaps the `counter` argument should be removed again?

oconnor663: Oh good catch, this new `counter` argument is from an earlier version of these changes (before I started relying more on `HasherExt`), and it doesn't make sense anymore.  I'll remove it.

> Shouldn't this be counter instead of 0?

The BLAKE3 root node always uses counter 0, either because it's the first and only chunk (index 0) or because it's a parent node (all parents use counter 0). So as you increment the counter on the XOF side of things, it's always equal to the output block index, regardless of the input length.

oconnor663: 11852d66cc15399aba1628857dbc4dc81a7b79d7

oconnor663:
Released as v1.8.0.

PR #461, opened 2025-04-02
Add -DBLAKE3_USE_TBB to pkg-config CFLAGS

silvanshade:
This PR adds `-DBLAKE3_USE_TBB` to the pkg-config CFLAGS.

Can you please try to include this before the next release with #460? Thanks.

@BurningEnlightenment @oconnor663

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -240,6 +240,8 @@ if(BLAKE3_USE_TBB)
       PUBLIC
         BLAKE3_USE_TBB)
   endif()
+  list(APPEND PKG_CONFIG_REQUIRES "tbb >= ${TBB_VERSION}")
+  list(APPEND BLAKE3_PKGCONFIG_CFLAGS -DBLAKE3_USE_TBB)
 endif()
 
 if(BLAKE3_USE_TBB)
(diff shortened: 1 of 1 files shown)


BurningEnlightenment (approved):
ugh, it would be nice if the world converged on a universal dependency-spec standard with **built-in** support in all major build tools.

Ericson2314:
> ugh, it would be nice if the world converged on a universal dependency-spec standard with built-in support in all major build tools.

FWIW, some CMake and Meson devs are jointly working on such a thing: https://cps-org.github.io/cps/index.html#

oconnor663:
I'll cut a release now. Out of curiosity, pushing a new tag to GitHub affect our callers that use CMake? Is there a dependency fetching mechanism that fetches the latest tag or anything like that?

BurningEnlightenment:
> Out of curiosity, pushing a new tag to GitHub affect our callers that use CMake? Is there a dependency fetching mechanism that fetches the latest tag or anything like that?

Nothing built-in but `vcpkg` & `conan` only update their ports to official tags. And of course there are probably a bunch of home-grown `FetchContent` based solutions which are free to do anything.

oconnor663:
Released as v1.8.1.

BurningEnlightenment:
> FWIW, some CMake and Meson devs are jointly working on such a thing: https://cps-org.github.io/cps/index.html#

@Ericson2314 Back when CPS was first submitted to the ISO mailing list in 2018, I sort of expected that the initiative would go nowhere and fizzle out, but apparently they added experimental support[^1] in CMake 3.31 and published a [blog post](https://www.kitware.com/navigating-cmake-dependencies-with-cps/) about it this week. Though meson seems to still hold off on implementing. Interesting indeed.

<sub>Now we only have to wait 6+ years until all LTS Distros shipping incompatible build-systems are out of support 🙃</sub>

[^1]: https://cmake.org/cmake/help/latest/command/install.html#package-info

PR #463, opened 2025-04-03
Fix pkg-config with multiple CFLAGS

silvanshade:
This PR fixes the pkg-config CFLAGS when multiple features are enabled.

The problem is that CMake expands `PKG_CONFIG_CFLAGS` as a semi-colon separated list but this only occurs if there are multiple entries.

Unfortunately this didn't show up for me because `-DBUILD_SHARED_LIBS` was not enabled.

The fix is to re-use the same logic as for the `Requires` field to manually expand the field before substitution.

I've also added some CI tests which should catch some of this in the future.

Apologies for the churn around all of this.

@BurningEnlightenment @oconnor663

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -258,8 +258,8 @@ jobs:
     - run: docker run -v $(pwd):/io -w /io messense/cargo-xwin cargo xwin test --target x86_64-pc-windows-msvc --features=mmap,rayon,traits-preview,serde,zeroize
 
   # Currently only on x86.
-  c_tests:
-    name: C tests SIMD=${{ matrix.simd }} TBB=${{ matrix.use_tbb }}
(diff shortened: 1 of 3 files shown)


BurningEnlightenment (approved):
*sigh* maybe I should've gone with a config header, propagating compiler switches across build-systems really isn't fun

silvanshade:
Yeah... thanks for merging.

I noticed there's one more thing that should be changed though: the C++ standard library should be included in `pkg-config` libs, and it should be the same used for building for ABI compatibility. Was going to try and push that here but since it's already merged I'll add that in another PR.

BurningEnlightenment:
> Was going to try and push that here but since it's already merged I'll add that in another PR.

Okay, I recommend putting the PR in Draft status to avoid this. Anyways, thanks for the contribution!

PR #464, opened 2025-04-03
Add C++ std lib to pkg-config

silvanshade:
We should also include the C++ standard library used for building the TBB component into the pkg-config file.

This PR adds that and appropriate tests for CI.

This check for the C++ standard library should work on Linux and macOS with `libc++` (LLVM) and `stdlibc++` (GNU).

For Windows (MSVC specifically) it doesn't do anything. A case for handling that platform could probably be added but as far as I know pkg-config isn't commonly used there and I'm not even sure how to set that up.


silvanshade:
@Ericson2314 will this approach work okay for the Nix configuration? Just want to make sure I'm not missing anything.

Ericson2314:
@silvanshade better test it, but I think it should work.

I've seen some debate of best practices for this sort of thing, however. See https://github.com/mesonbuild/meson/issues/14234 and linked issues.

BurningEnlightenment:
@silvanshade feel free to un-draft after you confirmed that the changeset is sufficient.

silvanshade:
@Ericson2314 Seems to work works as expected on Linux and macOS from the nix build system: https://github.com/silvanshade/nix/tree/blake3-tbb-fixed-pkg-config

@BurningEnlightenment This should be ready.

BurningEnlightenment (approved):

PR #466, opened 2025-04-05
Prepare for const

nazar-pc:
This is extracted from https://github.com/BLAKE3-team/BLAKE3/pull/439.

`arrayref` usage was reduced due to API being non-`const` and honestly a bit confusing to use. With a few trivial macros it can be removed in many places (I only touched those relevant to https://github.com/BLAKE3-team/BLAKE3/pull/439) and expanded code is much easier to read.

Portable implementation was refactored a bit to make functions `const fn`-friendly, but not marked as `const fn` just yet.

Hopefully these changes are not too controversial.

### DIFF
--- a/src/platform.rs
+++ b/src/platform.rs
@@ -1,5 +1,4 @@
 use crate::{portable, CVWords, IncrementCounter, BLOCK_LEN};
-use arrayref::{array_mut_ref, array_ref};
 
 cfg_if::cfg_if! {
     if #[cfg(any(target_arch = "x86", target_arch = "x86_64"))] {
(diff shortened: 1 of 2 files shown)


nazar-pc:
@oconnor663 do you think this can be merged as is?

I'd like to rebase https://github.com/BLAKE3-team/BLAKE3/pull/439 on it later.

I'm using domain separation for hashes in various places and using blake3 hash of a fixed string as a key, which would be nice to do with constants rather than hardcoding values for readability.

nazar-pc:
@oconnor663 any concerns here?

I'd really like to somehow make progress on `const` support and hopefully https://github.com/BLAKE3-team/BLAKE3/issues/478 soon after as well.

nazar-pc:
Can this be merged, please? There isn't really anything complex or such that changes MSRV here.

Issue #467, opened 2025-04-08
MinGW + TBB build doesn't work


Ericson2314:
CMake purposely excludes it. Can this be fixed? I did build TBB with MinGW.

CC @silvanshade.

silvanshade:
I believe it was excluded because at the time I wasn't able to get it to build with MinGW, but I don't remember the details.

If it is working though then we should support that.
 
Can you provide a little more information about how you were able to compile it? What versions of toolchain components were using (and how were they installed) and any relevant configuration details would help.

I can try to take a look and see if I can replicate a successful build from that.

Ericson2314:
With the recent Nixpkgs PR I merged (don't have link cause phone sorry) one can cross compile TBB for Windows with MinGW. Then when I tried to build Blake I got the error message. If you try to cross compile Blake with mingw in Nixpkgs master I think you should reproduce the issue straight-away.

silvanshade:
I tried to compile on Windows using two different pre-built toolchains recommended on https://www.mingw-w64.org/downloads/

- [w64devkit](https://github.com/skeeto/w64devkit)
- [msys2](https://www.msys2.org/)

I failed to compile with them which I detail below.

I haven't had a chance to test cross-compiling to MinGW with nix but I'm a little surprised it works.

I don't really have the bandwidth at the moment to try and track down the issues upstream in TBB and fix them though.

Maybe it would be best just to remove the check and add a note to the README about known working configurations. That would at least unblock the Nix packaging issues.

# w64devkit

Configuration:
```
cmake --fresh -S c -B c/build -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=c/build/install -DBLAKE3_USE_TBB=ON -DBLAKE3_EXAMPLES=ON -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++

CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.


-- The C compiler identification is GNU 14.2.0
-- The CXX compiler identification is GNU 14.2.0
-- The ASM compiler identification is GNU
[...]

PR #468, opened 2025-04-09
Remove compiler check for TBB feature

silvanshade:
Compilation may work with other configurations such as MinGW under the right circumstances.

Related: #467

CC @Ericson2314

@BurningEnlightenment @oconnor663 Would it be possible to make a new patch release with this and #464? That would help resolve some downstream issues for Nix:

- https://github.com/NixOS/nix/issues/12889#issuecomment-2787107051
- https://github.com/NixOS/nix/pull/12676


Ericson2314 (approved):

Mic92:
Otherwise simple patches like this can be also applied on top of tbb until the next release in nixpkgs.

BurningEnlightenment (approved):

PR #469, opened 2025-04-09
Check x86 features even in `no_std`

nazar-pc:
This makes it possible to use accelerated versions with runtime feature detection in `no_std` environment.

This is nice in projects where bare-metal implementation is needed or when most all the code is already `no_std` and requiring `std` just to get a faster version of blake3 is inconvenient.

I went with well known and well maintained `cpufeatures`, though eventually previously used macros should become usable from `::core`: https://github.com/rust-lang/rfcs/pull/2725

Note that `cpufeatures` already handles cases where features are enabled at compile time, so there is no need for explicit `#[cfg(target_feature = "X")]` blocks anymore.

The only thing that still depends on `std` feature now is `std::io`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -114,11 +114,14 @@ arrayref = "0.3.5"
 arrayvec = { version = "0.7.4", default-features = false }
 constant_time_eq = { version = "0.3.1", default-features = false }
 cfg-if = "1.0.0"
-digest = { version = "0.10.1", features = [ "mac" ], optional = true }
+digest = { version = "0.10.1", features = ["mac"], optional = true }
 memmap2 = { version = "0.9", optional = true }
 rayon-core = { version = "1.12.1", optional = true }
 serde = { version = "1.0", default-features = false, features = ["derive"], optional = true }
 zeroize = { version = "1", default-features = false, optional = true }
+
+[target.'cfg(any(target_arch = "x86", target_arch = "x86_64"))'.dependencies]
+cpufeatures = "0.2.17"
 
 [dev-dependencies]
 hmac = "0.12.0"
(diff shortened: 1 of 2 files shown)


oconnor663:
Interesting, this is my first time looking at the `cpufeatures` crate. I'll need to read the docs.

oconnor663:
Do you know what this line means [in their docs](https://docs.rs/cpufeatures/latest/cpufeatures/#supported-target-architectures)?

> NOTE: target features with an asterisk are unstable (nightly-only) and subject to change to match upstream name changes in the Rust standard library.

`avx512f` and `avx512vl` both have stars next to their names.

nazar-pc:
There was a drama around AVX10 versions with both 512-bit and 256-bit vectors, but eventually Intel decided to roll it back and [settled on 512-bit only](https://www.phoronix.com/news/Intel-AVX10-Drops-256-Bit). I think it is related to that, see https://github.com/rust-lang/rust/issues/138843

I don't know if `avx512f` and `avx512vl` will ever go away, but I'd speculate that probably not because there are existing CPUs that don't support any version of the AVX10, but do support some of those instructions and there must be a way to detect that.

I just tested on AMD 7970X CPU (Zen 4, AVX512-capable) and performance both before and after is the same, meaning the features are detected the same way.

Maybe @tarcieri can share some more details about this if he has time.

nazar-pc:
I think this is a better link, AVX512 features are about to be stabilized: https://github.com/rust-lang/rust/pull/138940

nazar-pc:
I'd appreciate 1.8.3 release with this change included, it has been a while since PR was merged

oconnor663:
Good idea, done!

PR #474, opened 2025-04-14
Added Clickhouse to the adoption list

shiv4289:
[Clickhouse uses BLAKE3](https://clickhouse.com/docs/sql-reference/functions/hash-functions#blake3)

It shows here in their [cargo.toml file](https://github.com/ClickHouse/ClickHouse/blob/master/rust/chcache/Cargo.toml#L7).

And they seem to be happy about it too

`BLAKE3 performs more than 2x faster than SHA224 or SHA256 and a bit faster than MD5. Also, BLAKE3 is secure, unlike MD5 and SHA-1 and protected against length extension, unlike SHA-2.`

More about their usage here - https://clickhouse.com/blog/more-than-2x-faster-hashing-in-clickhouse-using-rust


oconnor663:
Thanks!

PR #476, opened 2025-04-17
Remove `std` feature from `rayon` feature

nazar-pc:
I found no good reason for `rayon` feature to require `std`. In fact it'd be nice to just use `rayon` all the time sometimes regardless if it is used in `std` or `no_std` environment.

UPD: Looks like even `rayon-core` depends on `std` internally (which may or may not change in the future), but there is still no direct need for it in BLAKE3 itself.

This complements https://github.com/BLAKE3-team/BLAKE3/pull/469

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ std = []
 #
 # Implementation detail: We take a dependency on rayon-core instead of rayon,
 # because it builds faster and still includes all the APIs we need.
-rayon = ["dep:rayon-core", "std"]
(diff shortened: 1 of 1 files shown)


nazar-pc:
cc @oconnor663, not sure if you're getting notifications about Rust changes without explicit pings

oconnor663:
I think you're right on both counts. I don't think it's possible to pull in `rayon` without also pulling in `std`, but also there's no particularly reason we need to do this explicitly, and it's cleaner not to. Thanks!

PR #477, opened 2025-04-23
Relax Clang frontend variant detection

silvanshade:
This allows Clang frontend variant detection to work better in more scenarios such as when compiling with cross-compilers with platform prefixes.

In particular this fixes cross-compiles from Linux to FreeBSD for nix packaging.

It would be good to get this in the next patch release but probably isn't critical enough to warrant it's own release since it should only show up when using both Clang and specially named cross-compiler (which is somewhat unusual since Clang already supports multiple targets).

CC @Ericson2314

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -23,6 +23,7 @@ option(BLAKE3_FETCH_TBB "Allow fetching oneTBB from GitHub if not found on syste
 include(CTest)
 include(FeatureSummary)
 include(GNUInstallDirs)
+include(BLAKE3/Utils)
 
 add_subdirectory(dependencies)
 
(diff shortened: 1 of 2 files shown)


BurningEnlightenment:
Would the cross compiling detection automatically work if we used `CMAKE_CXX_COMPILER_FRONTEND_VARIANT`?

silvanshade:
> Would the cross compiling detection automatically work if we used `CMAKE_CXX_COMPILER_FRONTEND_VARIANT`?

Yes.

With that in mind I refactored the existing detection code to use `CMAKE_CXX_COMPILER_FRONTEND_VARIANT` if available and fall back to our re-implementation otherwise.

I also moved the logic into `BLAKE3/Utils.cmake` which is probably cleaner if you're moving some of the other stuff there in the other PR.

silvanshade:
@BurningEnlightenment can we merge this one also?

c/cmake/BLAKE3/Utils.cmake
```diff
@@ -0,0 +1,21 @@
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.10)
+  include_guard(GLOBAL)
+endif()
+
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.14)
+  set(BLAKE3_CXX_COMPILER_FRONTEND_VARIANT ${CMAKE_CXX_COMPILER_FRONTEND_VARIANT})
```
BurningEnlightenment, line 6:
```suggestion
  set(BLAKE3_CXX_COMPILER_FRONTEND_VARIANT "${CMAKE_CXX_COMPILER_FRONTEND_VARIANT}")
```

c/CMakeLists.txt
```diff
@@ -258,32 +259,19 @@ if(BLAKE3_USE_TBB)
     set(APPEND BLAKE3_CXX_STANDARD_FLAGS_GNU -std=c++20)
     set(APPEND BLAKE3_CXX_STANDARD_FLAGS_MSVC /std:c++20)
   endif()
+
   set(BLAKE3_CXXFLAGS_GNU "-fno-exceptions;-fno-rtti;${BLAKE3_CXX_STANDARD_FLAGS_GNU}" CACHE STRING "C++ flags used for compiling private BLAKE3 library components with GNU-like compiler frontends.")
   set(BLAKE3_CXXFLAGS_MSVC "/EHs-c-;/GR-;${BLAKE3_CXX_STANDARD_FLAGS_MSVC}" CACHE STRING "C++ flags used for compiling private BLAKE3 library components with MSVC-like compiler frontends.")
```
BurningEnlightenment, line 264: I've been thinking on whether it might make sense to merge the cache variables into a single `BLAKE3_CXXFLAGS` with a default based on `BLAKE3_CXX_COMPILER_FRONTEND_VARIANT`, but haven't had time to experiment.

BurningEnlightenment (approved):

c/cmake/BLAKE3/Utils.cmake
silvanshade: Unfortunately this PR is already referenced in the downstream patches to fix the BLAKE3 build and the PR contents is hashed in the package.

So I'll have to open a new PR with these changes: https://github.com/BLAKE3-team/BLAKE3/pull/483

c/CMakeLists.txt
silvanshade: What is the intended use case?

Why don't we do that in a separate PR just to keep things focused?

PR #482, opened 2025-04-29
build(cmake): Use tbb32 pkgconfig package on 32-bit builds

timschumi:
This now directly mirrors [the TBB-sided logic](https://github.com/uxlfoundation/oneTBB/blob/v2021.11.0/src/tbb/CMakeLists.txt#L157-L161) to determine the name of the pkg-config file.


silvanshade:
Thanks for fixing this.

@BurningEnlightenment maybe it would be good to add a 32-bit target build test in #472?

timschumi:
> maybe it would be good to add a 32-bit target build test in 472?

fwiw, I don't think CI would have found this unless one of the test cases is linking _against libblake3_ using the `pkgconfig` file (since that file is what is currently broken).

BurningEnlightenment (approved):
Thanks. Out of interest: Why do people write stuff this way? Is it a pkg-config deficiency? It looks incredibly brittle.

silvanshade:
> fwiw, I don't think CI would have found this unless one of the test cases is linking _against libblake3_ using the `pkgconfig` file (since that file is what is currently broken).

I did propose a test case for that in the referenced PR -- basically linking the example (with both pkg-config and then CMake), specifically to try and catch package configuration issues. But I wouldn't have thought the 32-bit case would need special handling without this fix you found.

PR #483, opened 2025-04-29
Relax Clang frontend variant detection (refactor of #477)

silvanshade:
This allows Clang frontend variant detection to work better in more scenarios such as when compiling with cross-compilers with platform prefixes.

This version includes refactorings from feedback in https://github.com/BLAKE3-team/BLAKE3/pull/477


silvanshade:
@BurningEnlightenment please review this PR in place of #477.

Also, would it be possible to configure the repository to be able to request reviews?

BurningEnlightenment (approved):

PR #485, opened 2025-05-22
`impl Deref for Hash`

nazar-pc:
Something I found needing over and over again.

Allows to index hash with `hash[0]`, slice `&hash[..4]`, convert to `[u8; 32]` with `*hash`.

The regular methods are still useful because they are `const fn`, but `Deref` makes things so much easier in many cases.

I wasn't sure if you'd like `DerefMut`, so I didn't add it (should be a rare need anyway).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -144,6 +144,7 @@ use arrayref::{array_mut_ref, array_ref};
 use arrayvec::{ArrayString, ArrayVec};
 use core::cmp;
 use core::fmt;
+use core::ops::Deref;
 use platform::{Platform, MAX_SIMD_DEGREE, MAX_SIMD_DEGREE_OR_2};
 #[cfg(feature = "zeroize")]
 use zeroize::Zeroize;
@@ -241,9 +242,18 @@ fn counter_high(counter: u64) -> u32 {
 #[derive(Clone, Copy, Hash, Eq)]
 pub struct Hash([u8; OUT_LEN]);
 
+impl Deref for Hash {
+    type Target = [u8; OUT_LEN];
+
+    #[inline(always)]
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
 impl Hash {
     /// The raw bytes of the `Hash`. Note that byte arrays don't provide
-    /// constant-time equality checking, so if  you need to compare hashes,
+    /// constant-time equality checking, so if you need to compare hashes,
     /// prefer the `Hash` type.
     #[inline]
     pub const fn as_bytes(&self) -> &[u8; OUT_LEN] {


oconnor663:
Did you see this comment in [the `Hash` docs](https://docs.rs/blake3/latest/blake3/struct.Hash.html)?

> Hash doesn’t implement [Deref](https://doc.rust-lang.org/stable/std/ops/trait.Deref.html) or [AsRef](https://doc.rust-lang.org/std/convert/trait.AsRef.html), to avoid situations where a type conversion happens implicitly and the constant-time property is accidentally lost.

I've never gotten any concrete evidence one way or the other about whether the tradeoff is worth it, but that's why it is the way it is. What do you think?

nazar-pc:
Hm... I certainly did not see that note. It is a noble goal to offer constant-time comparison, but I'm not sure `Hash` is really the right place.

If I was worried about constant-time, I'd do constant-time comparison explicitly rather than relying on a piece of documentation in the library.

Same with `zeroize`, it is not clear to me in what contexts hash of any value would be a security-sensitive thing on its own.

I value type safety, so `Hash` as a new type makes a lot of sense, but I'd personally limit it to that + convenient conversion to/from bytes and probably not much else.

What I ended up doing in current project is creating a bunch of new types around hash values for different contexts. `Hash` doesn't work for me because it doesn't implement serialization traits from a few libraries, doesn't have domain-specific methods attached to it, etc.

Looking at `Hash` even methods like `to_hex()` and `from_hex()` while useful, feel a bit out of place. Speaking of which, it is often the case that JSON serialization of hashes is a hex string, not a decimal array. I even implemented that in a `Blake3Hash` new type myself to ensure nice JSON and efficient binary serialization: https://github.com/autonomys/subspace/pull/3184/files#diff-4cc5e991fd987263e3a9d7ad96be8b024c79613ff4bbcae37da23aa1488b596aR50-R90

oconnor663:
> it is not clear to me in what contexts hash of any value would be a security-sensitive thing on its own

You might already know all of this this, but the central example is a `keyed_hash` output, which is similar to HMAC. A recipient who's trying to authenticate some message is going to recompute the keyed hash and see if that matches what the sender sent. If not, the message is inauthentic or corrupt, and it needs to be discarded. Crucially but often confusingly, if the comparison failed then the correct hash needs to remain secret. That's why the comparison needs to be constant-time (always), and it's also why you might want to zeroize that value (if you generally zeroize sensitive values).

nazar-pc:
I understand that you sometimes compare the hash, but the hash is the output. There are cases when you need constant time, but assuming you can't make BLAKE3 output the hash, even guessing the hash as such will not help you in many cases. This is a nuanced topic for sure and I hope people designing such systems themselves really know what they are doing and better yet use well designed off-the-shelf libraries rather than rolling their own crypto.

cesarb:
It's important to make it easier to do the right thing, and harder to do the wrong thing. When comparing cryptographic hashes for equality, having the comparison be constant-time is often a requirement. It's better to have the naive `hash(input) == expected` be constant-time, which in the worst case is just slightly slower, than having it not be constant-time and requiring every user to remember to do something like `constant_time_eq(&hash(input), expected)` when necessary. If you have a need to directly access the bytes of the hash, the `.as_bytes()` escape hatch is there for that purpose.

cesarb:
> Sure, but I guess I disagree that it is an important property of the hash output type. You _really_ should be using constant-time functions explicitly when you care about this kind of stuff. And for countless use cases constant-time is not a requirement at all.

For these use cases, you can do `hash(input).as_bytes() == expected`, and the only consequence of forgetting the `.as_bytes()` would be the code being slightly slower. On the other hand, for the use cases where using a constant-time comparison is important (and they might not be obvious, especially if you're not into cryptography), forgetting the `constant_time_eq()` or equivalent will open a timing vulnerability.

That is, we could have:

- `hash(input) == expected` (not constant time)
- `constant_time_eq(&hash(input), expected)` (constant time)

Or what we currently have:

- `hash(input) == expected` (constant time)
- `hash(input).as_bytes() == expected` (not constant time)

In that first alternative (in which `hash(input)` directly returns a `[u8; 32]`), forgetting to do the `constant_time_eq` (or equivalent) will silently do the wrong thing. In the second alternative (the current implementation), forgetting to do the `.as_bytes()` is just a minor performance loss (if any); doing the simplest thing (`hash(input) == expected`) is always correct.

PR #487, opened 2025-06-01
Fix mmap() conditions

nabijaczleweli:
Get the file size from seeking to the end (canonical, equivalent to File::stream_len()) instead of fstat(); this is more real and unseekable files are unmappable (conversely, in the vast majority of cases, seekable files are mappable)

Always try to map (size permitting) instead of only mapping S_IFREGs and fall back to reading on failure (instead of hard-erroring). This means we can map devices and correctly fall back if the mapping failed for any number of inocuous reasons (mapping too big, too many mappings, whatever)

This also ends up faster (lseek()+mmap() instead of statx()+statx()+mmap())

Modelled after https://git.sr.ht/~nabijaczleweli/voreutils/commit/e4621ad93ae69d28c032932ad617c026790c0a40

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -1,5 +1,8 @@
 //! Helper functions for efficient IO.
 
+#[cfg(feature = "mmap")]
+use std::io::Seek;
+
 #[cfg(feature = "std")]
 pub(crate) fn copy_wide(
     mut reader: impl std::io::Read,
@@ -21,9 +24,7 @@ pub(crate) fn copy_wide(
     }
 }
 
-// Mmap a file, if it looks like a good idea. Return None in cases where we know mmap will fail, or
-// if the file is short enough that mmapping isn't worth it. However, if we do try to mmap and it
-// fails, return the error.
+// Mmap a file, if it looks like a good idea. Return None if we can't or don't want to.
 //
 // SAFETY: Mmaps are fundamentally unsafe, because you can call invariant-checking functions like
 // str::from_utf8 on them and then have them change out from under you. Letting a safe caller get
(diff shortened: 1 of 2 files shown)


oconnor663:
Could you give me an example of something that gets mmap'ed now that wasn't getting mmap'ed before? Is there something under `/proc` on Linux?

The original `memmap` (not `memmap2`) had what was arguably a bug, where it could map files that were larger than `isize::MAX`, which is in fact UB in Rust. I'm pretty sure `memmap2` errors in that case, but if we're supplying the length explicitly I want to double check that we're still on a codepath that checks that for us.

oconnor663:
In fact I think `memmap2` is _not_ checking that `isize::MAX` limit for us if we set `len` explicitly: https://github.com/RazrFalcon/memmap2-rs/blob/91bc53aa13605c9538b94564d1c02cce424ad9c0/src/lib.rs#L241-L281. I'm a little bit surprised by that, and I'm considering filing it as a bug.

nabijaczleweli:
The main driver for this is devices; `b3sum /dev/rbd3` works now.

Why would it check that? This is a valid map:
```
lseek(3, 0, SEEK_END)                   = 2247483658
mmap(NULL, 2247483658, PROT_READ, MAP_SHARED, 3, 0) = 0x71449000
```
(this is 2^31+100000010).

It's not possible to make a file bigger than 2^63-1 (but even if it weren't, the map would just fail as usual when you give it a length that's too big).

nabijaczleweli:
CI failure appears unrelated

nabijaczleweli:
https://github.com/RazrFalcon/memmap2-rs/pull/145 merged

nabijaczleweli:
...and now rebuilding this picks up memmap2 0.9.7 which has the check implemented

oconnor663:
Apologies for the long delay here. I'm finally coming back to this question, and here's the approach I'm considering: https://github.com/BLAKE3-team/BLAKE3/pull/568. Let me know if you have any comments on that?

nabijaczleweli:
That doesn't fix reading devices.

oconnor663:
Oops, now I actually understand the problem. Here's another approach: https://github.com/BLAKE3-team/BLAKE3/pull/570. This might be trying to be "too clever" with the end-16383 seek, but I think it can save us a rewind syscall in the common case of hashing a short file.

nabijaczleweli:
Delicious, wfm :)

oconnor663:
Landed https://github.com/BLAKE3-team/BLAKE3/pull/570. Thanks for your help!

oconnor663:
Released as https://github.com/BLAKE3-team/BLAKE3/releases/tag/1.8.6.

PR #497, opened 2025-07-16
Mark blake3_hash_neon as static

nikic:
This is an implementation detail, which should not be exported.

This avoids unnecessary symbol conflicts with vendored copies of blake3.


BurningEnlightenment:
how do you handle the other symbol conflicts like those caused by the assembly implementations?

nikic:
> how do you handle the other symbol conflicts like those caused by the assembly implementations?

By prefixing the symbols: https://github.com/llvm/llvm-project/blob/main/llvm/lib/Support/BLAKE3/llvm_blake3_prefix.h

We could prefix this one as well, but making it static seems cleaner...

PR #506, opened 2025-08-09
bump constant_time_eq version to 0.4.2

silverstillisntgold:
Simple change to take advantage of the internal [improvements](https://github.com/cesarb/constant_time_eq/blob/0.4.2/CHANGES#L48) made to the constant_time_eq dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,9 @@ wasm32_simd = []
 # enabled by default. (Previously this also controlled runtime CPU feature
 # detection on x86, but now we use the no-std-compatible `cpufeatures` crate
 # for that.)
-std = []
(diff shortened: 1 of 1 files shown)


silverstillisntgold:
Checks currently fail on MSRV support, so this will have to be delayed until blake3 moves to the 2024 edition of rust (version 1.85).

oconnor663:
Yes I've been delaying bumping `constant_time_eq` for that reason. If there was a concrete benefit to some BLAKE3 benchmark, then I wouldn't wait this long, but I assume it's more of a "nice to have" bump that won't be showing up in the profiles of anyone doing a lot of hashing.

oconnor663:
I think it's time to take this. Thanks for pointing out the `std` feature!

PR #519, opened 2025-09-16
CMakeLists.txt: include CXX in project languages only when using TBB

stefan-sedlak:
Avoids unnecessarily enabling the C++ toolchain when TBB support is disabled.


stefan-sedlak:
Hi @BurningEnlightenment , could you please merge it? Or shall I improve something?

PR #520, opened 2025-09-20
chore: Set `rust-version` field for `blake3` and `b3sum`

sorairolake:
The `rust-version` field defines the MSRV of the crate. I think this is especially useful for library crates.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ license = "CC0-1.0 OR Apache-2.0 OR Apache-2.0 WITH LLVM-exception"
 documentation = "https://docs.rs/blake3"
 readme = "README.md"
 edition = "2021"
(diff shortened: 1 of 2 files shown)


oconnor663:
My habit has been not to set this when it's just a maximum of the MSRV of dependencies (as opposed to some recent language/stdlib feature that this crate is using itself). It seems like it tends to become stale over time as dependencies publish new versions. Is there a "community consensus" about how to manage this?

sorairolake:
> My habit has been not to set this when it's just a maximum of the MSRV of dependencies (as opposed to some recent language/stdlib feature that this crate is using itself).

OK. I close this pull request.

> Is there a "community consensus" about how to manage this?

To my knowledge, there is no such community consensus.

PR #521, opened 2025-09-29
add explicit unsafe blocks to more unsafe functions

oconnor663:
These blocks silence warnings that we'll see when we eventually turn on the 2024 edition. I used Claude to make these changes.

I'm putting up this PR to test automated Codex reviews. First time...


oconnor663:
@codex if you had to nitpick something what would it be?

PR #527, opened 2025-10-14
Use `panic = "abort"` in `b3sum`

kotauskas:
As `b3sum` does not attempt to recover from panics (nor does it have any reason to attempt to do that), it has no use for Rust's default behavior of unwinding on panics. Switching `b3sum` to `panic = "abort"` makes the executable about 13% smaller (from 1.5 MiB to 1.3 MiB) and possibly provides more optimization opportunities to LLVM. Note that this does not impact the display of backtraces, which are still displayed using `libunwind` – the only features that are disabled are the ability to catch panics using `std::panic::catch_unwind` and the non-fatal nature of panics in non-main threads.

### DIFF
--- a/b3sum/Cargo.toml
+++ b/b3sum/Cargo.toml
@@ -24,3 +24,11 @@ wild = "2.0.3"
 [dev-dependencies]
 duct = "1.0.0"
 tempfile = "3.1.0"
+
+# b3sum does not gracefully handle panics (nor does it make sense for it to do
+# that), so unwinding on panic would just be a minor performance loss and
+# binary size hog with no benefit
+[profile.dev]
+panic = "abort"
+[profile.release]
+panic = "abort"


oconnor663:
That makes sense to me. What's your thinking around doing it in dev + release vs just in release? (Is there any chance we'll miss some extra debug info or something when tests panic at failed asserts?)

(CI failures are clearly unrelated, [upstream issue](https://github.com/RustCrypto/traits/issues/2041).)

kotauskas:
> What's your thinking around doing it in dev + release vs just in release?

Mostly just consistency, but it might improve compile times a tiny bit.

> Is there any chance we'll miss some extra debug info or something when tests panic at failed asserts?

[Tests always use unwinding panics](https://doc.rust-lang.org/cargo/reference/profiles.html#panic) even if you have `panic = "abort"` in `[profile.dev]`. As for the unwind tables, they seem to get generated either way, just without the unneeded landing pads.

kotauskas:
@oconnor663 gentle reminder about this trivial PR

PR #528, opened 2025-10-21
chore: make some documents clearer

wyrapeseed:
make some documents clearer

### DIFF
--- a/c/blake3.h
+++ b/c/blake3.h
@@ -38,7 +38,7 @@ extern "C" {
 #define BLAKE3_MAX_DEPTH 54
 
 // This struct is a private implementation detail. It has to be here because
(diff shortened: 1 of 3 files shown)


b3sum/src/main.rs
```diff
@@ -25,6 +25,7 @@ struct Inner {
     /// Files to hash, or checkfiles to check
     ///
     /// When no file is given, or when - is given, read standard input.
+    /// The - argument represents standard input.
```
oconnor663: Thanks for the PR. I like the other changes, but this one feels a little redundant with the line above. What do you think?

oconnor663:
(CI failures are a known upstream issue, unrelated.)

oconnor663:
Thanks!

PR #532, opened 2025-11-17
Add `Hash::as_slice()` for convenient serialization to bytes

joshtriplett:
`Hash::as_bytes()` returns the hash as an array of bytes. However,
sometimes it's useful to have the hash as a slice of bytes instead, such
as for serialization via an API that takes `&[u8]`. While
`.as_bytes().as_slice()` works for this, it'd be more convenient to have
a direct `.as_slice()` on hashes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -253,6 +253,14 @@ impl Hash {
         Self(bytes)
     }
 
+    /// The raw bytes of the `Hash`, as a slice. Useful for serialization. Note that byte arrays
+    /// don't provide constant-time equality checking, so if you need to compare hashes, prefer
+    /// the `Hash` type.
+    #[inline]
+    pub const fn as_slice(&self) -> &[u8] {
+        self.0.as_slice()
+    }
+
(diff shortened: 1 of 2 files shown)


oconnor663:
(Ah ignore my last comment, I misread this as `AsRef`.)

Is there a situation where the `.as_bytes()` array won't coerce to a slice? I guess if you're trying to use it in a generic context like `impl std::io::Read`?

joshtriplett:
@oconnor663 Yeah, generic contexts are exactly the problem here. For instance, passing `.as_slice()` to a database's API for `INSERT`, which takes a generic trait that supports byte slices but not byte arrays.

joshtriplett:
Anything I could do to help with this?

joshtriplett:
Thank you!

oconnor663:
Thanks for the PR. Do you need a prompt release?

joshtriplett:
> Thanks for the PR. Do you need a prompt release?

Not a prompt one; I'll use it when released but I have a local helper working around it so it isn't urgent. Would appreciate one at your convenience.

PR #541, opened 2026-02-04
c: fix cmake - force c++20 only when c++ is used

pol51:
We don't need to force C++20 when no C++ code in compiled in BLAKE3 (so it's only needed when BLAKE3_USE_TBB=ON)


stefan-sedlak:
Related / Duplicate https://github.com/BLAKE3-team/BLAKE3/pull/519

PR #549, opened 2026-04-13
c: Use correct SIMD flags when compiling with Clang-Cl

QianNangong:
Avoid AVX code when compiling with Clang-Cl


BurningEnlightenment (changes requested):
I'd prefer a nested if-statement to avoid duplicating `BLAKE3_AMD64_ASM_SOURCES`.

Btw. and just out of curiosity: Are you compiling for x86 and if yes, why?

QianNangong:
> I'd prefer a nested if-statement to avoid duplicating `BLAKE3_AMD64_ASM_SOURCES`.

Fixed.
> Btw. and just out of curiosity: Are you compiling for x86 and if yes, why?

Yes, I'm using blake3 inside a legacy project to replace MD5 checksum, and it should be able to run on a 32-bit Windows 7 environment.

BurningEnlightenment (approved):

PR #560, opened 2026-06-17
Add  `rand` feature that implements RNG traits for `OutputReader`

Le-Maz:
## Summary

This PR adds an optional `rand` feature that implements following traits for `blake3::OutputReader`:
- `rand::TryRng`
- `rand::TryCryptoRng`
- `rand::rand_core::block::Generator`

Implementing this I tried to follow the pattern of the existing optional `zeroize` feature.

## Motivation

I believe it would be natural for `OutputReader` to implement the `rand` traits so it can be used directly as a random number generator after `finalize_xof()`.

Example usage:
```rs
use blake3::Hasher;
use rand::{Rng, RngExt};

hasher.update(b"seed data");
let mut reader = hasher.finalize_xof();

let my_num = reader.next_u64();
let my_range = reader.random_range(1..100);
```

This change is a minimal interop for these crates. While the separate `rand_blake3` crate exists, it appears to be orphaned (no updates since late 2023, and the Git repository is deleted) and implements more than just the core traits on `OutputReader`. Therefore I propose a very small, focused implementation, so that it's not a hassle to have in there.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,9 @@ mmap = ["std", "dep:memmap2"]
 # Implement the zeroize::Zeroize trait for types in this crate.
 zeroize = ["dep:zeroize", "arrayvec/zeroize"]
 
+# Implement the rand::TryRng trait for OutputReader.
+rand = ["dep:rand"]
+
 # This crate implements traits from the RustCrypto project, exposed here as the
 # "traits-preview" feature. However, these traits aren't stable, and they're
 # expected to change in incompatible ways before they reach 1.0. For that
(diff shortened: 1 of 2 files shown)


elichai:
I'm not sure if this is a good idea as long as OutputReader is not buffered

Le-Maz:
I understand the performance penalty, but I wanted to keep the proposed changes minimal. I think buffering can be offloaded to generic wrappers as a separation of concerns.

But thinking of the change again, I probably should add impl TryCryptoRng as well.

oconnor663:
It probably makes sense to go ahead and make the larger change of doing buffering in `OutputReader`. We added Unix-only AVX-512 support on the output side [a couple years ago](https://github.com/BLAKE3-team/BLAKE3/releases/tag/1.5.4). We meant to expand that to Windows/AVX/ARM, but we haven't done that yet. I think what we have today justifies the feature though.

We'd need a 16*64 bytes = 1 KiB buffer to support AVX-512. That's pretty big, but `Hasher` is bigger, and I only remember one instance where anyone complained about that. (I think the C version triggered a stack size warning when someone tried using it in the Linux kernel?)

I apologize for being much less responsive with code reviews this year. Baby life is no joke :)

Le-Maz:
Okay then, I've made an issue about the buffering (#561) and restated some of what you wrote.

PR #562, opened 2026-06-28
b3sum: add system libblake3 backend

catap:
Many distributions ship libblake3.so as a separate installed package, see: https://repology.org/project/blake3/versions , so allow b3sum to link against that library with the system feature, rather than always relying on an embedded C build.

The default Rust backend remains unchanged.

With the system feature enabled, b3sum uses the C bindings as its Rust side API surface, while the bindings build script asks pkg-config for libblake3 and links the installed library instead of compiling the bundled C sources.

### DIFF
--- a/b3sum/src/main.rs
+++ b/b3sum/src/main.rs
@@ -19,6 +19,39 @@ const RAW_ARG: &str = "raw";
 const TAG_ARG: &str = "tag";
 const CHECK_ARG: &str = "check";
 
+#[cfg(feature = "system")]
+use blake3_c_rust_bindings as blake3;
+
+#[cfg(not(feature = "system"))]
+type OutputReader = blake3::OutputReader;
+
+#[cfg(feature = "system")]
+#[derive(Clone)]
+struct OutputReader {
+    hasher: blake3::Hasher,
+    position: u64,
+}
+
+#[cfg(feature = "system")]
+impl OutputReader {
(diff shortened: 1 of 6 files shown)


oconnor663:
> rather than always relying on an embedded C build

`b3sum` uses the Rust `blake3` crate, never the `blake3.c` implementation. You might've noticed that the `build.rs` in `blake3` includes some low-level C files on some platforms by default (`blake3_neon.c` on AArch64) or not by default (`blake3_avx512.c` on x86_64), but that's the extent of the sharing. Also see this note in `c/blake3_c_rust_bindings/README.md`:

> these bindings are not expected to be used in production. They're intended for testing and benchmarking.

Currently I don't want to change this.

b3sum/src/main.rs
```diff
@@ -173,25 +206,72 @@ impl Args {
...
             bail!("Cannot open `-` in keyed mode");
         }
-        hasher.update_reader(io::stdin().lock())?;
+        update_reader(&mut hasher, io::stdin().lock())?;
     } else if args.no_mmap() {
-        hasher.update_reader(File::open(path)?)?;
+        update_reader(&mut hasher, File::open(path)?)?;
     } else {
-        // The fast path: Try to mmap the file and hash it with multiple threads.
+        #[cfg(not(feature = "system"))]
```
oconnor663, line 257: Note that this would make `--features=system` a huge performance regression for most users.

Issue #565, opened 2026-07-26
Proposal to change Rust default from hand-written assembly to native Rust intrinsics


nazar-pc:
It bothered me for a while that Rust version was slower than assembly, so I wanted to get to the bottom of it. Thankfully, it is now fairly quick and easy with LLMs, and results ended up a bit surprising.

Turns out Rust version produces meaningfully different assembly, which, depending on compilation target, can be both slower and substantially faster.

Testing was done by benchmarking on a fixed CPU core on AMD Threadripper 7970X (Zen 4). Each time I did 3 runs of SSE2, SSE4.1 and AVX2 builds.

Assembly build (3 runs on SSE2, SSE4.1 and AVX2):
```
> RUSTFLAGS="" taskset -c 1 cargo +nightly bench --bench bench --features no_avx512,no_avx2,no_sse41 -- atonce_1024 2>&1 | grep bench_atonce_1024_kib
test bench_atonce_1024_kib             ... bench:     362,618.50 ns/iter (+/- 4,147.63) = 2891 MB/s
test bench_atonce_1024_kib             ... bench:     362,664.10 ns/iter (+/- 3,123.62) = 2891 MB/s
test bench_atonce_1024_kib             ... bench:     364,221.90 ns/iter (+/- 10,890.57) = 2878 MB/s

> RUSTFLAGS="" taskset -c 1 cargo +nightly bench --bench bench --features no_avx512,no_avx2 -- atonce_1024 2>&1 | grep bench_atonce_1024_kib
test bench_atonce_1024_kib             ... bench:     340,970.20 ns/iter (+/- 1,460.55) = 3075 MB/s
test bench_atonce_1024_kib             ... bench:     339,213.05 ns/iter (+/- 2,814.00) = 3091 MB/s
test bench_atonce_1024_kib             ... bench:     341,099.80 ns/iter (+/- 3,012.18) = 3074 MB/s

> RUSTFLAGS="" taskset -c 1 cargo +nightly bench --bench bench --features no_avx512 -- atonce_1024 2>&1 | grep bench_atonce_1024_kib
test bench_atonce_1024_kib             ... bench:     176,468.54 ns/iter (+/- 7,651.41) = 5942 MB/s
test bench_atonce_1024_kib             ... bench:     175,342.40 ns/iter (+/- 1,626.39) = 5980 MB/s
[...]

PR #566, opened 2026-07-27
Add Rust AVX-512 intrinsics implementation

nazar-pc:
This is a more or less direct port of AVX-512 C implementation to Rust. When compiled for Zen 4 it is slightly faster than C implementation and noticeably slower than assembly:
```
Assembly:
test bench_atonce_1024_kib             ... bench:     120,048.27 ns/iter (+/- 912.41) = 8734 MB/s
C:
test bench_atonce_1024_kib             ... bench:     129,066.54 ns/iter (+/- 5,228.13) = 8124 MB/s
Rust:
test bench_atonce_1024_kib             ... bench:     124,809.20 ns/iter (+/- 1,697.71) = 8401 MB/s
```

The reason for C and Rust being slower is register spilling, LLVM can't quite manage that many live values properly and spills zmm registers to the stack. I plan to create an LLVM issue about it and see what can be done about that (especially in context of https://github.com/BLAKE3-team/BLAKE3/issues/565).

I also removed AVX-512 support from 32-bit builds (arguably useless and just bloats the binary) and changed AVX-512 preference to Rust instead of C since performance is comparable. However, since AVX-512 intrinsics are only supported since Rust 1.89.0, older Rust versions will not compile Rust version of AVX-512 and fall back to existing C version instead. This way there is no change in supported Rust versions.

CI was adjusted slightly to not only test new code path, but also the fallback case.

UPD: LLVM issue: https://github.com/llvm/llvm-project/issues/212375

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,6 +163,12 @@ jobs:
     # the MSRV, but it's not a promise that we won't bump it.
     - uses: dtolnay/rust-toolchain@1.85.1
     - run: cargo build --features=mmap,rayon,traits-preview,serde,zeroize
+    # AVX-512 intrinsics were stabilized in Rust 1.89, after this MSRV. Build with
+    # "prefer_intrinsics" and "pure" to make sure build.rs correctly detects that and
+    # falls back (to the C intrinsics, and to no AVX-512 code at all, respectively)
+    # instead of trying to build src/rust_avx512.rs.
+    - run: cargo build --features=prefer_intrinsics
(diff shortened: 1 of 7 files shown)


nazar-pc:
I updated the implementation with similar changes to https://github.com/BLAKE3-team/BLAKE3/pull/573 and updated the benchmark numbers as well. There is still a gap, but it is much smaller than before and I'll try to find out the reason for the remaining performance difference one day, especially for https://github.com/BLAKE3-team/BLAKE3/issues/565 to be viable with AVX512.

PR #567, opened 2026-07-27
Fix path to Visual Studio toolchain in CI

nazar-pc:
Visual Studio was updated recently, this is why all recent PRs are failing CI. Something less fragile would have been nice to have, ideally with support for not just x86-64, but also aarch64 Windows runners.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -444,7 +444,7 @@ jobs:
           sudo apt-get install libtbb-dev libtbb12
       - name: CMake generation, build, install
         run: |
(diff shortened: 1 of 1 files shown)


oconnor663:
Thank you! Codex suggests using `vswhere` to this less fragile. What do you think of the version on this branch? https://github.com/BLAKE3-team/BLAKE3/tree/refs/heads/vswhere

nazar-pc:
Yeah, that looks much better and should not break again in the future

oconnor663:
Awesome, done.

PR #569, opened 2026-08-04
Add an SVE2 backend for AArch64 that fuses xor+rotate with XAR

BobDu:
SVE2 adds `XAR`, which computes `rotr(a ^ b, imm)` in one instruction. BLAKE3's G function does exactly that eight times per round, and on NEON each one takes two or three instructions (`EOR`, then `REV32`, `TBL`, or `SHL`+`SRI`), so `round_fn4` goes from 128 vector instructions to 80 at the same 128-bit width.

This adds `blake3_sve2.c`, a copy of `blake3_neon.c` with those pairs fused, opt-in via `BLAKE3_USE_SVE2` and off by default. It is built with `-msve-vector-bits=128` because every SVE2 core shipping today is 128 bits wide, and `blake3_dispatch.c` checks `HWCAP2_SVE2` and the runtime vector length before using it, falling back to NEON otherwise.

`cargo +nightly bench --features=neon,sve2 many_` in `c/blake3_c_rust_bindings`, single-threaded, MB/s:

| | chunks NEON | chunks SVE2 | parents NEON | parents SVE2 |
|---|---|---|---|---|
| Neoverse N2 (GitHub `ubuntu-24.04-arm`) | 1557 | 2227 (1.43x) | 1354 | 1790 (1.33x) |
| Neoverse V2 (Graviton4, `m8g.large`) | 1540 | 2032 (1.32x) | 1347 | 1684 (1.25x) |
| Neoverse V3 (Graviton5, `m9g.large`) | 1827 | 3271 (1.79x) | 1630 | 2666 (1.70x) |

The Rust crate still uses NEON on AArch64; if this is merged, I will wire SVE2 into it in a follow-up PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,6 +122,22 @@ jobs:
     - name: cargo test C bindings intrinsics
       run: cargo test --features=prefer_intrinsics
       working-directory: ./c/blake3_c_rust_bindings
+    # SVE2 is AArch64-only and needs its own -march flags, so this is restricted to
+    # the arm64 runner. GitHub's Arm-hosted runners are Neoverse N2 with a 128-bit
+    # vector length, so the backend is reached rather than falling back to NEON.
+    - name: cargo test C bindings SVE2
+      if: matrix.target.os == 'ubuntu-24.04-arm'
+      run: cargo test --features=sve2
+      working-directory: ./c/blake3_c_rust_bindings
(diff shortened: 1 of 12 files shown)


BobDu:
@oconnor663 I noticed you may not have much time for this repo at the moment, so I reworked this PR to be quicker to review:

it is now two commits, and the first one only copies `blake3_neon.c` to `blake3_sve2.c` byte for byte, so the second commit's diff is exactly what changes for SVE2 plus the dispatch, build and test wiring.

PR #573, opened 2026-08-08
Improve AVX2 AVX512 C/Rust performance

nazar-pc:
I used LLM to figure out why C and Rust implementations are slower than assembly and it turns out the cause is different algorithm. So here are the changes that bring C and Rust implementations more in line with assembly that recovers a substantial amount of performance difference.

For AVX512 the difference is ~10% higher performance than before. It is still lower, but much less so. Commit messages have some benchmarking results. I tested on Zen 4 CPU and results are also in the same overall direction even though absolute numbers are different.

The last commit fixes some warnings on nightly, the rest are caused by `cpufeatures` and will be addressed once https://github.com/RustCrypto/utils/pull/1515 is merged and released, so nightly failures are temporarily expected.

### DIFF
--- a/c/blake3_avx2.c
+++ b/c/blake3_avx2.c
@@ -232,7 +232,7 @@ INLINE void transpose_msg_vecs(const uint8_t *const *inputs,
   }
 
   for (size_t i = 0; i < 8; ++i) {
-    _mm_prefetch((const void *)&inputs[i][block_offset + 256], _MM_HINT_T0);
+    _mm_prefetch((const void *)&inputs[i][block_offset + 128], _MM_HINT_T0);
   }
 }
 
--- a/c/blake3_avx2.c
+++ b/c/blake3_avx2.c
@@ -4,9 +4,15 @@
 
 #define DEGREE 8
 
-INLINE __m256i loadu(const uint8_t src[32]) {
-  return _mm256_loadu_si256((const __m256i *)src);
-}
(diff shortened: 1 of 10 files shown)


nazar-pc:
Performance is A LOT better in C version with Clang rather than default GCC 15.2.0 on Ubuntu 26.04:
```
Assembly:
test bench_atonce_1024_kib             ... bench:     120,048.27 ns/iter (+/- 912.41) = 8734 MB/s
C (GCC 15.2.0 on Ubuntu 26.04):
test bench_atonce_1024_kib             ... bench:     129,066.54 ns/iter (+/- 5,228.13) = 8124 MB/s
C (Clang from https://github.com/llvm/llvm-project/commit/2e3553def14cbc819013c107ff3edd4fb08e348c):
test bench_atonce_1024_kib             ... bench:     123,441.73 ns/iter (+/- 1,642.13) = 8494 MB/s
C (Clang with some WIP patches I'm working on):
test bench_atonce_1024_kib             ... bench:     122,814.63 ns/iter (+/- 1,659.77) = 8537 MB/s
```

nazar-pc:
Prefetch distance was 256 in C/Rust for some reason, while 128 in assembly. Changing to 128 improved performance further slightly on Clang trunk:
```
test bench_atonce_1024_kib             ... bench:     122,536.34 ns/iter (+/- 1,470.58) = 8557 MB/s
```

PR #582, opened 2026-08-28
align the two on-stack block buffers to 64 bytes

poliebotics:
As invited in [zooko/bench-hashes#2](https://github.com/zooko/bench-hashes/issues/2#issuecomment-5415063851): this wraps `ChunkState.buf` and `Output.block` in a `repr(align(64))` newtype, `Aligned64`. glibc's AVX-512 memcpy writes a 64-byte copy as two overlapping unaligned zmm stores, and when the destination is not 32-byte-aligned, the immediate reload of the buffer fails store-to-load forwarding. Stack frame placement decides that alignment, so identical binaries land in distinct 64-byte-input timing bands per process. Aligning the buffers pins every process to the fast case. Full mechanism work (watchpoint traces, PMU counters, a hash-free reproducer) is in the linked issue.

Verification of the original alignment commit (f2b2278, captured 28 August 2026), all on Zen 5 (Ryzen AI MAX+ PRO 395), glibc 2.43, rustc 1.98.0, pinned core / FIFO / ASLR off / empty env, stack phase set by a trampoline (`and rsp,-64; sub rsp,SHIFT`). The CPU was capped at 3.0 GHz for this capture, so absolute ns run ~1.6x the issue's boost-clocked numbers while the band ratios reproduce:

**64 B phase sweep** (8 shifts x 3 interleaved reps, 20M iters/point), class medians in ns/op:

| | shift%64=0 | 16 | 32 | 48 |
|---|---|---|---|---|
| master | 94.16 | 134.54 (+42.9%) | 94.15 | 126.70 (+34.6%) |
| this PR | 94.04 | 94.05 | 94.08 | 94.06 |

Re-checked after the `Aligned64` rework (commit 0e4eda6, 10 September 2026, uncapped clocks, one rep, class medians): same shape, master +35.2% and +44.3% in its two slow classes, this PR's four classes within 2% of each other and its eight points spanning 55.85 to 58.42 ns against master's fast-phase 55.77 to 58.25 ns. The shift-112 point sits 4-5% high on both sides alike.

**STLF-failure counter** (`ls_bad_status2.stli_other` per hash): master 2.07-2.08 in the fast phases, 4.10 and 5.06 in the slow ones; this PR 2.02-2.03 in every phase. Retired instructions are flat across phases on both sides.

**Buffer residue** (gdb, break at `ChunkState::update`): master's `&chunk_state & 63` walks 0/48/32/16 as the frame shifts; this PR reads 0 at every shift.

**Long inputs** (1 KiB / 16 KiB / 64 KiB / 1 MiB, 4 phases x 5 interleaved reps): all median deltas within ±0.43%, and the patched build's cross-phase spread is tighter than master's at every length.

**Streaming** (split-update 80 B, a 1-byte then a 79-byte update, forcing the buffered-block compress path): flat on both sides at every tested shift in steady state, 268-271 ns/op, deltas within 0.9%; the patch's frozen frame sits at master's level.

**This repo's bench suite** (`cargo bench` via `RUSTC_BOOTSTRAP=1`, two interleaved full rounds per side): 67 benches, median delta +0.01%, everything within -1.6% to +0.8% except `bench_single_compression_portable`. Its two full-round medians differed by +3.3%, so I reran it filtered, three more executions per side: the five-execution medians are 71.33 ns (master) vs 71.18 ns (this PR). The two excursions in that capture (master 106.03 once, this PR 76.29 once) turned out to be a machine transient under its capped power profile rather than the code: six warmed back-to-back reruns read 71.15-71.51 on both sides.

**Tests**: a local stable-x86_64 mirror of ci.yml, 51 commands green on the current branch (0e4eda6, 10 September 2026): every library feature combo including `rayon,traits-preview,serde,zeroize`, the full SIMD-tier matrix (assembly / `prefer_intrinsics` / `pure`, debug and release, down to `no_sse2`), `--no-default-features`, `--benches`, `test_vectors` x3, `reference_impl`, all six C-bindings configs, `b3sum` (both test configs and the LTO build) and `cargo fmt --check`, plus the MSRV 1.85.1 build. There's a new `test_block_buffer_alignment` pinning the alignment and field-offset invariants.

Layout cost on x86_64 (unchanged by the `Aligned64` rework): `size_of::<Hasher>()` 1920 -> 1984, `size_of::<OutputReader>()` 120 -> 192, and the deprecated `guts::ChunkState` 112 -> 128, all now 64-byte-aligned. No public signatures change; the public types' Rust layout changes as above, and a downstream `#[repr(packed)]` container holding one of these types would stop compiling (packed types cannot transitively contain `repr(align)` fields). No unsafe.

Two notes for review. The issue's direct evidence implicates `ChunkState.buf`; aligning `Output.block` as well is the configuration I benchmarked end to end, and it pins the other hot 64-byte buffer's residue instead of leaving it frozen wherever the realigned frame drops it. The issue documents that hazard concretely: a build that aligned only `Output.block` froze the still-unaligned `ChunkState.buf` at a slow residue in every process, so the two alignments travel together here. Second, the C implementation: measured to a pre-registered decision rule and closed with no change proposed. No analogous timing band appears in any tested C path on this binary/chip/glibc: this gcc build inlines the 64-byte one-shot fill (glibc is never called there), and the paths that do pair a glibc fill with an immediate compress (a temporary chunk state's 64-byte fill in the multi-chunk branch, and the sub-64 boundary fill) show stall counts that vary with the buffer's frame-given residue while end-to-end timing stays flat within 1%. A whole-block-copy specialization at the 64-byte site is worth 0.062% at 1025 B (95% CI 0.03-0.10%, six paired rounds), which is not a change worth review time. Two rig facts for the record: the full eight-residue map confirms the issue's rule at n=64, slow exactly when the destination is not 32-byte-aligned, and 63-byte copies through this glibc stall at every residue, a uniform toll that alignment cannot address.

All timing is one chip and one glibc: I observed no regression beyond measurement noise in these captures, and the linked issue carries the cross-hardware caveats.

— BOSUN ⚓


zooko:
@oconnor663

src/lib.rs
```diff
@@ -432,10 +432,30 @@ impl std::error::Error for HexError {}
 // Each chunk or parent node can produce either a 32-byte chaining value or, by
 // setting the ROOT flag, any number of final output bytes. The Output struct
 // captures the state just prior to choosing between those two possibilities.
```
oconnor663: This comment needs to be moved now.

src/lib.rs
```diff
@@ -432,10 +432,30 @@ impl std::error::Error for HexError {}
...
+
+#[derive(Clone, Copy)]
+#[repr(align(64))]
+struct ChunkBuf([u8; BLOCK_LEN]);
+
 #[derive(Clone)]
 struct Output {
     input_chaining_value: CVWords,
-    block: [u8; 64],
+    block: OutBlock,
```
oconnor663: I see why you called this an "out block" here, but it's really more accurately described as an _input_ block. It's 64 bytes of compression function input (either the padded final block of a short message, or the root parent node bytes of a longer one), which we re-compress as many times as needed to produce output.

I think it would be simpler to unify `ChunkBuf` and `OutBlock` into a single, general-purpose `Aligned64` type, which we use in both places.

src/test.rs
```diff
@@ -809,6 +809,17 @@ const fn test_hash_const_conversions() {
     _ = hash.as_bytes();
 }
 
+#[test]
+fn test_block_buffer_alignment() {
+    // These buffers must stay 64-byte-aligned so that wide vector stores into
+    // them (e.g. from memcpy) can store-to-load forward regardless of stack
+    // layout; see the comment on OutBlock/ChunkBuf in lib.rs.
```
oconnor663: As we update the type names in this comment, let's also add links to the relevant issue thread with the benchmarks.

src/lib.rs
```diff
@@ -458,15 +478,20 @@ impl Output {
...
+            self.flags | ROOT,
+        );
         Hash(platform::le_bytes_from_words_32(&cv))
     }
 
     fn root_output_block(&self) -> [u8; 2 * OUT_LEN] {
         self.platform.compress_xof(
             &self.input_chaining_value,
-            &self.block,
+            &self.block.0,
```
oconnor663: As part of unifying on `Aligned64`, let's add relevant trait impls like `Deref` and `Eq`, so that most of these little tweaks disappear.

oconnor663 (changes requested):

oconnor663:
Thanks for spending time on this. (I get that an llm wrote the code, but I assume a human also spent time on it? :sweat_smile:) It looks good to me with the tweaks above.

poliebotics:
Hoy, and thanks for the review. Addressed in 0e4eda6. One `Aligned64` type now serves both buffers, with `Deref`/`DerefMut`, `PartialEq`/`Eq` (plus `PartialEq<[u8; 64]>`) and a `Zeroize` impl behind the feature, so the call sites and the zeroize assertions read as they did before this PR. The `Output` comment is back directly above `struct Output`, and the alignment test's comment names `Aligned64` and links the issue thread and this PR. I re-ran the 64 B sweep on the same rig, one rep at boost clocks. The baseline still shows its two slow phases, +35% and +44% today, while the reworked build spans 55.9 to 58.4 ns across the eight shifts, matching the baseline's own fast-phase range of 55.8 to 58.3 ns, and `&ChunkState & 63` reads 0 in the four residue classes inspected. The local ci.yml mirror is green again, 51 commands on stable x86_64 plus the 1.85.1 MSRV build. I kept this as a separate commit so the delta is easy to see. Happy to squash if you would rather merge one.

On the human question, yes. Cathal Ryan Hynes, whose account this is, commissioned the investigation, approved the mechanism write-up and this PR, and runs the pushes and posts, this one included. Someone also has to switch the power back on, which at his end is a daily event at an hour of its own choosing, so until that is fixed he gets co-author on everything I write. The code, the measurements and the prose are mine, an LLM-driven assistant working under his direction. We were already deep in BLAKE3 for ZeeBeam, his zero-knowledge work over projector-camera recordings, where each frame's Groth16 proof binds a BLAKE3 feedback chain, so your stack frames were a pleasant detour.

— BOSUN ⚓

PR #586, opened 2026-09-07
Fix build warning for unused variable

mchoo7:
Another way to fix this is `(void)num_cvs;` or tweaking the `#if` macro below, but this is the simplest approach.


mchoo7:
Upstreaming from https://github.com/llvm/llvm-project/pull/221779

mchoo7:
MSVC doesn't support `__attribute__`. Any recommendation for other approaches?

What I can do is:

```c
  size_t num_cvs = blake3_compress_subtree_wide(input, input_len, key,
                                                chunk_counter, flags, cv_array, use_tbb);
  assert(num_cvs <= MAX_SIMD_DEGREE_OR_2);
  (void)num_cvs;
```

